PR #5, opened 2021-07-16
Initial draft of `vetiver_dashboard()`

juliasilge:
Basically just the markdown template for now


inst/rmarkdown/templates/vetiver_dashboard/skeleton/skeleton.Rmd
```diff
@@ -0,0 +1,49 @@
+---
+title: "Monitor model"
+output: 
+  vetiver::vetiver_dashboard:
+    board: 'pins::board_rsconnect()'
```
cderv, line 5: Just to share that this will be passed as a string to the function. So `board` will be equal to `"pins::board_rsconnect()"` inside your format function `vetiver_dashboard`. 

This means that `board$versioned` will probably error. I think you need to evaluate the expression that you are passing as a string through YAML. 

Otherwise you could pass evaluated R expression but this would require `!expr`

````yaml
output: 
  vetiver::vetiver_dashboard:
    board: !expr pins::board_rsconnect()
````

And maybe it would create other issue 🤔 

**bslib** leverages that when you want to pass a `bs_theme()` object using YAML
https://rstudio.github.io/bslib/articles/bslib.html#low-level
It was the only possibility at the beginning but this has been simplified so that YAML key value are passed as argument to `bs_theme` instead

````yaml
output:
  html_document:
    theme:
      bg: "#002B36" 
      fg: "#EEE8D5"
      progress-bar-bg: "orange"
````

equivalent to
````yaml
output:
    html_document:
        theme: !expr bs_theme(bg = "#002B36", fg = "#EEE8D5",  "progress-bar-bg" = "orange")
)
````

So you could imagine something like 

````yaml
output: 
  vetiver::vetiver_dashboard:
    board: 
        board_rsconnect
               auth: envvar
              version: FALSE
```` 

Just some thoughts as I was looking at your PR for your initial rmarkdown question

File: R/dashboard.R [hunk 2]
```diff
@@ -0,0 +1,51 @@
...
+        pin_navbar <- list(
+            title = "Model Pin",
+            icon = "fa-location-dot",
+            href = v$metadata$url,
+            target = "_blank"
+        )
+        navbar <- append(navbar, list(pin_navbar))
+    }
+
+    flexdashboard::flex_dashboard(!!!dashboard_dots, navbar = navbar)
```
cderv, line 49: You probably need to use `rlang::inject()` here (https://rlang.r-lib.org/reference/inject.html) because you are using `!!!` operator in a function that does not have built-in support for it. 

**rlang** documents this as out of context usage https://rlang.r-lib.org/reference/topic-inject-out-of-context.html#using-and-out-of-context

```suggestion
    rlang::inject(flexdashboard::flex_dashboard(!!!dashboard_dots, navbar = navbar))
```

cderv (commented):
Hi @juliasilge ! 

Following your question about splicing dot list, I looked at your PR and took the opportunity to share some thoughts regarding your custom format creation and Rmd YAML. 

Happy to discuss further.

juliasilge:
Closed in #98

PR #18, opened 2021-10-14
Complete partial argument match

schloerke:
Fixes https://github.com/tidymodels/vetiver/issues/17

(Untested)


schloerke:
Tested. The warning goes away.

PR #19, opened 2021-10-14
Use `rlang::is_interactive()`

schloerke:
... I should also do this in `plumber`

(untested)

- [x] `document()`


juliasilge (approved):
Thanks! 🙌

PR #50, opened 2021-10-21
Try out new generic scheme

juliasilge:
This new generic scheme involves `vetiver_model()` being a plain function and instead the developer needing to write six 6️⃣ methods:

- four for creating the model object (two of these are "maybe" but the person needs to think through them)
- two for the API handlers

This is three more methods than before, but we have it more explicitly broken out by task.


juliasilge:
This works out pretty well for xgboost:

``` r
library(vetiver)
library(pins)
library(plumber)

model_board <- board_folder(path = "/tmp/test", versioned = TRUE)
car_xgb <- xgboost::xgboost(as.matrix(mtcars[,-1]), 
                            mtcars$mpg, 
                            nrounds = 3, 
                            objective = "reg:squarederror")
#> [1]  train-rmse:14.931315 
#> [2]  train-rmse:10.956807 
#> [3]  train-rmse:8.086657
v <- vetiver_model(car_xgb, "cars_xgb", model_board)
vetiver_pin_write(v)
#> Creating new version '20211021T223836Z-11cd9'
#> Writing to pin 'cars_xgb'

pr() %>%
    vetiver_pr_predict(v, debug = TRUE)
#> # Plumber router with 1 endpoint, 4 filters, and 0 sub-routers.
#> ├──[queryString]
#> ├──[body]
#> ├──[cookieParser]
#> ├──[sharedSecret]
#> └──/predict (POST)
 ## next pipe to pr_run(port = 8088) to see visual documentation
```

<sup>Created on 2021-10-21 by the [reprex package](https://reprex.tidyverse.org) (v2.0.1)</sup>

PR #56, opened 2021-11-03
add methods for mlr3

be-marc:
https://github.com/mlr-org/mlr3/issues/718

Deploy model.

```r
library(mlr3)
library(vetiver)

task = tsk("pima")
learner = lrn("classif.rpart")

learner$train(task)

v <- vetiver_model(learner, "pima_rpart")
v

library(plumber)

pr() %>%
  vetiver_pr_predict(v) %>%
  pr_run(port = 8088)
```
Predict with deployed model.

```r
library(mlr3)
library(plumber)
library(vetiver)
library(tibble)

task = tsk("pima")
newdata = as_tibble(task$data())[, task$feature_names]

endpoint <- vetiver_endpoint("http://127.0.0.1:8088/predict")
predict(endpoint, newdata)

# > # A tibble: 768 × 1
# >    diabetes
# >    <chr>   
# >  1 pos     
# >  2 neg     
# >  3 neg     
# >  4 neg     
# >  5 pos     
# >  6 neg     
# >  7 neg     
# >  8 neg     
# >  9 pos     
# > 10 pos     
# > # … with 758 more rows
```


juliasilge:
Thank you so much for this contribution @be-marc! 🙌 We appreciate it so much and this looks great.

I notice that you added a prototype to your trained model object (which is the same approach that Max took with caret, rather than passing in training data and wrangling that or something). From my perspective, this means I'll wait on the next CRAN release of vetiver until after the next CRAN release of mlr3. I don't have a specific time frame or rush in mind, but when generally do you expect to do that?

juliasilge:
Failure on Windows 3.6 here is due to brand new xgboost which is not building

PR #67, opened 2021-12-21
Modularize plumber functions

### LINKED ISSUE #65: Decouple API Spec changes from `/predict` endpoint definition
Currently a vetiver plumber file contains the following that defines the `/predict` endpoint:

```r
#* @plumber
function(pr) {
    pr %>% vetiver_pr_predict(v)
}
```

`vetiver_pr_predict()` does all of the heavy lifting:
* uses the appropriate handler to define the `/predict` endpoint
* re-writes the OpenAPI spec file to define the API input data and other details
* changes the docs rendering from Swagger to Rapidoc.

This proposal is to separate the endpoint definition from the OpenAPI tasks so that instead the definition might look something like the following:

```r
#* @plumber
function(pr) {
    pr %>% 
        vetiver_pr_predict(v) %>%
        vetiver_pr_openapi(v)
}
```

In this case, the user is free to define other endpoints in addition to `/predict` as well as completely remove the generated `/predict` endpoint if desired. The OpenAPI function will then inspect the router and adjust the spec file for every endpoint that identifies itself (via request method and arguments) as a model endpoint by providing requestBody definition. This way, users can define multiple model endpoints without needing to worry about defining their own spec for each endpoint.

juliasilge:
Closes #65


juliasilge:
This is in process as of right now. The more modular functions have been created, but the `api_spec()` function still only works on one endpoint and there is not yet sniffing out of appropriate endpoints to apply it to.

juliasilge:
I think this is a good default to go with for now; all the POST endpoints now update the API spec.

Try this:
``` r
library(vetiver)
library(plumber)
cars_lm <- lm(mpg ~ cyl + disp, data = mtcars)
v <- vetiver_model(cars_lm, "cars1")
v$metadata <- list(url = "potatoes")

another_handler <- function(req) {
    newdata <- req$body
    sum(newdata[names(v$ptype)])
}

p <- pr() %>%
    vetiver_pr_post(v) %>%
    pr_post(path = "/sum", handler = another_handler) %>%
    vetiver_pr_docs(v)

p
#> # Plumber router with 3 endpoints, 4 filters, and 0 sub-routers.
#> ├──[queryString]
#> ├──[body]
#> ├──[cookieParser]
#> ├──[sharedSecret]
#> ├──/pin-url (GET)
#> ├──/predict (POST)
#> └──/sum (POST)
```

<sup>Created on 2022-01-04 by the [reprex package](https://reprex.tidyverse.org) (v2.0.1)</sup>

Which gets you this:

Screen Shot 2022-01-04 at 1 27 00 PM
And after clicking through, this:
Screen Shot 2022-01-04 at 1 27 38 PM

juliasilge:
@blairj09 Do you want to take a look at this to see if you have any more feedback? I could make looking for all the POST endpoints optional, I guess, as an additional argument to `api_spec()`.

R/pr-predict.R
```diff
@@ -3,6 +3,10 @@
 #' Use `vetiver_pr_predict()` to add a POST endpoint for predictions from a
 #' trained, pinned [vetiver_model()] to a Plumber router.
 #'
+#' This function uses `vetiver_pr_post()` for endpoint definition and
+#' `vetiver_pr_docs` to create visual API documentation; these more modular
```
blairj09, line 7:
```suggestion
#' `vetiver_pr_docs()` to create visual API documentation; these more modular
```

blairj09 (approved):
This approach looks good to me!

Issue #64, opened 2021-12-09
Create new `vetiver_deploy_sagemaker()` function


juliasilge:
Much like [`vetiver_deploy_rsconnect()`](https://vetiver.tidymodels.org/reference/vetiver_deploy_rsconnect.html) but for... SageMaker!

We can use [this example as some guidance](https://sagemaker-examples.readthedocs.io/en/latest/r_examples/r_byo_r_algo_hpo/tune_r_bring_your_own.html) and use `create_endpoint()` and/or `create_model()` from [paws](https://paws-r.github.io/docs/sagemaker/).

For the container, I wonder if we should start with the [`rstudio/plumber`](https://hub.docker.com/r/rstudio/plumber/) base image or the base R image from AWS? I would think the first option?

juliasilge:
I have a working demo here now: https://github.com/juliasilge/vetiverdemo/blob/main/biv-svm/train_sagemaker.Rmd

We can definitely wrap some of that up nicely, but some (like the special `sm-build` bit) will probably be hard to do from R.

PR #71, opened 2022-01-07
Create a Docker file

juliasilge:
Related to #64


R/write-docker.R
```diff
@@ -0,0 +1,104 @@
...
+    copy_renv <- glue("COPY {lockfile} /opt/ml/renv.lock")
+    copy_plumber <- glue("COPY {plumber_file} /opt/ml/plumber.R")
+
+
+    ret <- compact(list(
+        "# Generated by the vetiver package; edit with care\n",
+        ## https://github.com/rstudio/plumber/blob/main/Dockerfile:
+        "FROM rstudio/plumber:latest",
+        r_version,
+        "ENV R_VERSION=$R_VERSION\n",
```
juliasilge, line 52: I think this does not work, actually, to get the correct version of R into the build.

What is my best option here? Build from an image like `rocker/r-ver` and basically copy the guts of the `rstudio/plumber` Dockerfile?

tests/testthat/_snaps/write-docker.md
```diff
@@ -0,0 +1,61 @@
+# create plumber.R with packages
+
+    Code
+      cat(readr::read_lines(file.path(tmp_dir, "Dockerfile")), sep = "\n")
+    Output
+      # Generated by the vetiver package; edit with care
+      
+      FROM rstudio/plumber:latest
+      ARG R_VERSION=<r_version>
+      ENV R_VERSION=$R_VERSION
```
juliasilge, line 10: Pretty sure this doesn't work 😕

schloerke: Correct. :-(

Two options I can think of:
1. Build the `rstudio/plumber` Docker image and save it as a new custom name. Then use the custom name in the vetiver image. This would reduce installations of plumber, but I'm not too worried about that.
2. Make a combo Docker image as `rstudio/plumber` is not that big can could be handled by `pak`. 

I like second solution. I'll comment on the earlier conversation about implementation

R/write-docker.R
schloerke: Yes. `rstudio/plumber` Dockerfile is very light weight. I don't believe it is too bad to copy over some of the logic. I knowingly provide `plumber`'s Dockerfile with as minimal infrastructure as possible to encourage discovery, but not really permanent usage.

Thoughts:
* You are correct, we should point to `FROM rocker/r-ver:${R_VERSION}` as the parent image
* If we add `remotes` to https://github.com/tidymodels/vetiver/blob/d565de2035cea3a8bd605ca2f7a46d59aa9315b1/R/write-plumber.R#L83, then it'll get auto installed.
	* Do you need `renv` to be installed from GitHub? It might just be easier to add it to the `infra_pkgs`.
* You are already querying package manager, so all system deps should be auto-installed. ☑️ 
* We would need to add an updated entry point for the file:
```dockerfile
EXPOSE 8000
ENTRYPOINT ["R", "-e", "pr <- plumber::plumb(rev(commandArgs())[1]); pr$run(host = '0.0.0.0', port = 8000)"]
```
This is good, as the `rstudio/plumber` ENTRYPOINT was setting the docs to `TRUE`, which would overwrite your custom docs.

juliasilge: Thank you so much @schloerke! If you have time to take a look at the new snapshots tests, that would be fantastic.

PR #72, opened 2022-01-27
Add `check_ptype` arg

juliasilge:
This PR separates the act of saving a data prototype (`save_ptype` when creating a model) with using that prototype for checking at prediction time (`check_ptype`). The default behavior stays the same but this lets a user add a prototype for visual documentation but not use it for the vetiver checking (just use whatever the model does natively with wrongly typed data).

This change is at least partly to maintain parity with the Python implementation for vetiver.

@isabelizimm can you take a look at this and see if this aligns with what we spoke about?

### DIFF
--- a/R/pr-predict.R
+++ b/R/pr-predict.R
@@ -6,6 +6,9 @@
 #' @param pr A Plumber router, such as from [plumber::pr()].
 #' @param vetiver_model A deployable [vetiver_model()] object
 #' @param ... Other arguments passed to `predict()`, such as prediction `type`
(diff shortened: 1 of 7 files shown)


R/pr-predict.R
```diff
@@ -70,7 +73,8 @@ vetiver_pr_post <- function(pr,
                             vetiver_model,
                             path = "/predict",
                             debug = is_interactive(),
-                            ...) {
+                            ...,
+                            check_ptype = TRUE) {
```
isabelizimm, line 77: This looks great! The `check_ptype` arg will have the same experience on the python side 👍

Issue #73, opened 2022-02-02
Consider creating Plumber APIs with a `QUERY` endpoint in addition to `POST`


atheriel:
## Background

At my previous job we served model predictions via Plumber APIs (with identical endpoint names, even). Originally these endpoints used `GET` requests, but eventually some models required arguments too long or complex to put into a query string, so we moved to using `POST` with the arguments in the body. The reason for this is that `GET` requests are not expected to have a body, and many programs or intermediate proxies will strip them if you try. This is also how `vetiver` works today, if I am reading it correctly, and I assume it is for similar reasons.

Unfortunately, `POST` is a bad semantic fit for model predictions. Predictions don't have side effects and are idempotent, so they are 100% safe to retry if a request fails and could theoretically be cached at any intermediate layer.

However, the HTTP specification explicitly marks `POST` requests [as non-cacheable and non-idempotent by default](https://datatracker.ietf.org/doc/html/rfc7231#section-4.3.3), which has some relevant consequences:

* Clients (like browsers) must always re-request the prediction if they need it later.
* Prediction requests cannot supply "conditional" headers like those used for `GET` requests -- e.g. `If-Modified-Since` -- that could potentially reduce a lot of computation on the server side.
* This is true for anything in between as well, e.g. reverse proxies like NGINX cannot cache `POST` requests.
* Many HTTP request libraries will quietly retry failed requests, but they won't do this for `POST` requests by default because the results could be unpredictable.

You might think that caching is unimportant, but in my experience quite a lot of prediction requests are for a small set of domain-specific "base cases".

Moreover, because `vetiver` is *generating* an API that most users probably won't customise, we'd ideally like it to be a really performant and well-behaved one out of the box.

## Proposed Feature

There is a widely-supported addition to the HTTP specification currently under review at the IETF that [will add a new `QUERY` method](https://www.ietf.org/id/draft-ietf-httpbis-safe-method-w-body-02.html):

> This specification defines the HTTP QUERY request method as a means of making a safe, idempotent request that contains content.

`QUERY` supports caching and conditional headers, and unlike `GET` the URI is not expected to represent a resource.

It's my belief that this new method is a perfect fit for usecases like model prediction, so I'd propose that we jump on this bandwagon early and support a `QUERY` endpoint for generated Plumber APIs in `vetiver`.

## Potential Issues

[...]

PR #74, opened 2022-02-04
Deprecate `vetiver_pr_predict()` in favor of using `vetiver_pr_post()` directly

juliasilge:
We are consider whether it is too confusing to have `predict` methods for a vetiver model / endpoint and to also have a function called `vetiver_pr_predict()` so I have this option here:

``` r
library(vetiver)
library(plumber)
pr() %>% vetiver_pr_post(v)
#> ├──[queryString]
#> ├──[body]
#> ├──[cookieParser]
#> ├──[sharedSecret]
#> ├──/ping (GET)
#> └──/predict (POST)
```

<sup>Created on 2022-02-04 by the [reprex package](https://reprex.tidyverse.org) (v2.0.1)</sup>

I actually don't like this at all, though, now that I have implemented it. This function actually adds 2 (or 3) endpoints, only one of which is POST. As discussed in #73 it's possible we use an entirely different method at some point in the future. 😕

@blairj09 I am wondering if you have an opinion on this. Do you think it's confusing to have `vetiver_pr_predict()` along with those `predict` methods? Any suggestions?


juliasilge:
Not going with this option (on account of how I HATE IT) 😅

PR #76, opened 2022-02-09
Add vetiver support for `ranger::ranger()`

juliasilge:
Support for ranger, including options like quantile regression:

``` r
library(vetiver)
library(pins)
library(plumber)

cars_rf <- ranger::ranger(mpg ~ ., data = mtcars, quantreg = TRUE)
v <- vetiver_model(cars_rf, "cars_ranger", ptype_data = mtcars[,-1])

model_board <- board_folder(path = "/tmp/test")
vetiver_pin_write(model_board, v)
#> Replacing version '20220209T222207Z-506ea' with '20220209T224840Z-2f6df'
#> Writing to pin 'cars_ranger'
#> 

pr() %>%
    vetiver_pr_predict(v, debug = TRUE, type = "quantiles")
#> ├──[queryString]
#> ├──[body]
#> ├──[cookieParser]
#> ├──[sharedSecret]
#> ├──/ping (GET)
#> └──/predict (POST)
## next pipe to pr_run(port = 8088) to see visual documentation
```

<sup>Created on 2022-02-09 by the [reprex package](https://reprex.tidyverse.org) (v2.0.1)</sup>


juliasilge:
@isabelizimm When a user has to pass in some data to create an input data prototype, I set this up as `ptype_data`. That's consistent with what you have, right? Like this?

```r
v <- vetiver_model(cars_rf, "cars_ranger", ptype_data = mtcars[,-1])
```

isabelizimm:
> @isabelizimm When a user has to pass in some data to create an input data prototype, I set this up as `ptype_data`. That's consistent with what you have, right? Like this?
> 
> ```r
> v <- vetiver_model(cars_rf, "cars_ranger", ptype_data = mtcars[,-1])
> ```

That is correct! Users will have an identical experience for that argument.

PR #77, opened 2022-02-12
Deprecate `vetiver_pr_predict()` in favor of `vetiver_api()`

juliasilge:
We are consider whether it is too confusing to have `predict` methods for a vetiver model / endpoint and to also have a function called `vetiver_pr_predict()` so I have this option here:

``` r
library(vetiver)
library(plumber)
pr() %>% vetiver_api(v)
#> ├──[queryString]
#> ├──[body]
#> ├──[cookieParser]
#> ├──[sharedSecret]
#> ├──/ping (GET)
#> └──/predict (POST)
```

<sup>Created on 2022-02-11 by the [reprex package](https://reprex.tidyverse.org) (v2.0.1)</sup>

I like this way better than #74


juliasilge:
@isabelizimm This is working way better than the other option and is growing on me.

PR #86, opened 2022-04-07
Customize rapidoc

juliasilge:
Eventually add the vetiver hex


R/api.R
```diff
@@ -110,9 +110,17 @@ vetiver_pr_docs <- function(pr,
...
+         width=60px style=\"margin-left:7px\"/>'
     pr <- plumber::pr_set_docs(
-        pr, "rapidoc",
-        heading_text = paste("vetiver", utils::packageVersion("vetiver"))
+        pr,
+        "rapidoc",
+        slots = logo,
+        heading_text = paste("vetiver", utils::packageVersion("vetiver")),
+        header_color = "#0F4B71",
+        primary_color = "#407483"
```
juliasilge, line 123: @schloerke I wonder if you have any ideas to help me here with how to we can get the vetiver logo into the plumber docs. If someone has vetiver installed, they can get the little file I added via `system.file("vetiver.png", package = "vetiver")` but how do I get that into the "logo" slot so that it shows up correctly?

I can see how to add something that is available via https, like I did in this example here, but how do I add a static file? I looked [at the rapidoc setup](https://github.com/meztez/rapidoc/blob/master/R/paths.R) but that is all at one path that is passed as `static` to `plumber::register_docs()`.

schloerke: Two approaches I can think of...

* Add your own static directory
	* Something like `pr$mount("/vetiver/logo", system.file("LOGO/FOLDER", package = "vetiver"))`
	* Then the `` `src` should contain a relative path to `"../vetiver/logo/LOGO.png"` (Must traverse from `/docs/` back to `/`)
* Host the vetiver logo from gh-pages and just pull in the latest logo, similar to how they have done above.
	* Can use an address like `https://vetiver.tidymodels.org/dev/logo.png`

juliasilge:
What this looks like right now:

RapiDoc_matchy

I think I like that!

schloerke:
So cool! Great styling!

PR #90, opened 2022-04-20
Specify port when creating Dockerfile

### LINKED ISSUE #82: Allow user to specify port when creating Dockerfile
The port really shouldn't be hardcoded into the Dockerfile, and for example [Google Cloud Run has this as an env variable](https://cloud.google.com/run/docs/container-contract#port). h/t @jnolis

juliasilge:
Closes #82 

This first bit at least gets the port into the function arguments (no longer hard-coded).


juliasilge:
This now does support listening on a port that is injected as an environment variable (like on [Google Cloud Run](https://cloud.google.com/run/docs/container-contract#port)). You would do:

```r
vetiver_write_docker(v, port = 'as.numeric(Sys.getenv("PORT"))')
```

I took out `EXPOSE THE_PORT_NUMBER` from the generated Dockerfiles because it seems like [this isn't really important for the typical use case](https://stackoverflow.com/a/47594352/5468471) we have and it's a blocker for using the port from an environment variable. Hopefully folks with more advanced use cases can add it themselves.

juliasilge:
Do you have any suggestions on this @jnolis?

jnolis:
This looks good to me! My only possible concern would be that having arbitrary expressions for the port might have unintended side effects, but I think the scenario where you don't have a fixed port is rare enough that people who use that functionality should be able to handle the risks.

juliasilge:
Thank you so much! ✨

PR #92, opened 2022-05-10
Three monitoring functions (for use in dashboards)

juliasilge:
This PR adds three functions:

-  `vetiver_compute_metrics()`
-  `vetiver_pin_metrics()`
-  `vetiver_plot_metrics()`


juliasilge:
The three functions in action:

``` r
library(vetiver)
library(dplyr)
#> 
#> Attaching package: 'dplyr'
#> The following objects are masked from 'package:stats':
#> 
#>     filter, lag
#> The following objects are masked from 'package:base':
#> 
#>     intersect, setdiff, setequal, union
library(parsnip)
data(Chicago, package = "modeldata")
Chicago <- Chicago %>% select(ridership, date, one_of(stations))
training_data <- Chicago %>% filter(date < "2008-12-31")
testing_data <- Chicago %>% filter(date > "2009-01-01", date < "2010-12-31")
monitoring <- Chicago %>% filter(date > "2011-01-01", date < "2012-12-31")
lm_fit <- linear_reg() %>% fit(ridership ~ ., data = training_data)

library(pins)
b <- board_temp()

## before starting monitoring, initiate the metrics and pin
## (for example, with the testing data):
original_metrics <-
    augment(lm_fit, new_data = testing_data) %>%
    vetiver_compute_metrics(date, "week", ridership, .pred, .every = 4L) %>%
    vetiver_pin_metrics(date, b, "lm_fit_metrics", initiate = TRUE)
#> Guessing `type = 'rds'`
#> Creating new version '20220510T233158Z-3f4d9'
#> Writing to pin 'lm_fit_metrics'
original_metrics
#> # A tibble: 81 × 5
#>    date           n .metric .estimator .estimate
#>    <date>     <int> <chr>   <chr>          <dbl>
#>  1 2009-01-02     6 mae     standard       3.74 
#>  2 2009-01-02     6 rmse    standard       4.42 
#>  3 2009-01-02     6 rsq     standard       0.561
#>  4 2009-01-08    28 mae     standard       2.98 
#>  5 2009-01-08    28 rmse    standard       4.61 
#>  6 2009-01-08    28 rsq     standard       0.576
#>  7 2009-02-05    28 mae     standard       1.16 
#>  8 2009-02-05    28 rmse    standard       1.90 
#>  9 2009-02-05    28 rsq     standard       0.916
#> 10 2009-03-05    28 mae     standard       0.929
#> # … with 71 more rows

## to continue monitoring with new data, compute metrics and update pin:
new_metrics <-
    augment(lm_fit, new_data = monitoring) %>%
    vetiver_compute_metrics(date, "week", ridership, .pred, .every = 4L) %>%
    vetiver_pin_metrics(date, b, "lm_fit_metrics")
#> Guessing `type = 'rds'`
#> Replacing version '20220510T233158Z-3f4d9' with '20220510T233158Z-63557'
#> Writing to pin 'lm_fit_metrics'
new_metrics
#> # A tibble: 162 × 5
#>    date           n .metric .estimator .estimate
#>    <date>     <int> <chr>   <chr>          <dbl>
#>  1 2009-01-02     6 mae     standard       3.74 
#>  2 2009-01-02     6 rmse    standard       4.42 
#>  3 2009-01-02     6 rsq     standard       0.561
#>  4 2009-01-08    28 mae     standard       2.98 
#>  5 2009-01-08    28 rmse    standard       4.61 
#>  6 2009-01-08    28 rsq     standard       0.576
#>  7 2009-02-05    28 mae     standard       1.16 
#>  8 2009-02-05    28 rmse    standard       1.90 
#>  9 2009-02-05    28 rsq     standard       0.916
#> 10 2009-03-05    28 mae     standard       0.929
#> # … with 152 more rows

library(ggplot2)
vetiver_plot_metrics(new_metrics, date) +
    scale_size(range = c(2, 4))
```



<sup>Created on 2022-05-10 by the [reprex package](https://reprex.tidyverse.org) (v2.0.1)</sup>

DESCRIPTION
```diff
@@ -42,6 +42,9 @@ Suggests:
     callr,
     caret,
     covr,
+    curl,
+    dplyr,
```
juliasilge, line 46: This PR adds A LOT of packages here. I put all the packages for monitoring in `Suggests` because I am up against 20 packages in Imports already and I can sort of make the argument that people may want to deploy a model but not monitor it. I am very open to ideas for reducing this somehow.

juliasilge: I'm only using dplyr in examples and tests FWIW 🤷‍♀️

R/monitor.R
```diff
@@ -0,0 +1,151 @@
...
+#'     vetiver_pin_metrics(date, b, "lm_fit_metrics")
+#'
+#' library(ggplot2)
+#' vetiver_plot_metrics(new_metrics, date) +
+#'     scale_size(range = c(2, 4))
+#'
+#' @export
+vetiver_compute_metrics <- function(data,
+                                    date_var,
+                                    .period,
```
DavisVaughan, line 74: Addressed in rewrite.

I don't think you actually need to put a `.` in front of the slider args. This was only done because the `...` in `slide_period_dfr()` could be named and we want to avoid conflicts with what the user could put in. But you control what is passed through here, and there are no conflicts.

R/monitor.R
```diff
@@ -0,0 +1,151 @@
...
+                                    truth, estimate, ...,
+                                    metric_set = yardstick::metrics,
+                                    .every = 1L,
+                                    .origin = NULL,
+                                    .before = 0L,
+                                    .after = 0L,
+                                    .complete = FALSE) {
+
+    rlang::check_installed("slider")
+    metrics_dots <- list2(...)
```
DavisVaughan, line 84: Addressed in rewrite.

Capturing with `list2()` isn't going to work if you have class probability metrics like `roc_auc()`. These require that the user pass more bare column names through `...`, i.e. like `roc_auc(hpc_cv, truth = obs, VF, L, M)`.

Capturing with `list2()` will try and evaluate them immediately, which is too early. They need to be captured with `enquos()` OR you can pass the `...` all the way through to where you call `metric_set()`, which is what I ended up doing.

R/monitor.R
```diff
@@ -0,0 +1,151 @@
...
+    date_var <- enquo(date_var)
+    slider::slide_period_dfr(
+        data,
+        .i = data[[quo_name(date_var)]],
```
DavisVaughan, line 88: Addressed in rewrite

I think it will be more robust to give users the full tidyselect capabilities here, by using `tidyselect::eval_select()`. You then just have to ensure that it only selects 1 column. I've added `eval_select_one()` to do this which is a small wrapper I've used in a few functions now.

R/monitor.R
```diff
@@ -0,0 +1,151 @@
...
+        .period = .period,
+        .f = ~ tibble::tibble(
```
DavisVaughan, line 90: Addressed in rewrite

I've pulled this out into its own function, `compute_metrics()` to be a little clearer

R/monitor.R
```diff
@@ -0,0 +1,151 @@
...
+            !!date_var := min(.x[[quo_name(date_var)]]),
```
DavisVaughan, line 91: Addressed in rewrite

I think you will be happier in the long term if this is always given the same name, i.e. maybe use `.index` as the column name.

I've also renamed `n` to `.n`. All the columns that come back from `metric_set()` have a `.` in front of them, so I figure we should be consistent.

R/monitor.R
```diff
@@ -0,0 +1,151 @@
...
+}
+
+#' @rdname vetiver_compute_metrics
+#' @export
+vetiver_pin_metrics <- function(df_metrics,
+                                date_var,
+                                board,
+                                metrics_pin_name,
+                                initiate = FALSE) {
+    date_var <- quo_name(enquo(date_var))
```
DavisVaughan, line 111: Could use `eval_select_one()` here

R/monitor.R
```diff
@@ -0,0 +1,151 @@
...
+        new_metrics <- vec_sort(df_metrics)
+    } else {
+        new_dates <- unique(df_metrics[[date_var]])
+        old_metrics <- pins::pin_read(board, metrics_pin_name)
+        old_metrics <- vec_slice(
+            old_metrics,
+            ! old_metrics[[date_var]] %in% new_dates
+        )
+        new_metrics <- vec_sort(vctrs::vec_rbind(old_metrics, df_metrics))
+    }
```
DavisVaughan, line 123: Yea the fact that this is mostly one big if/else block really makes me think that the `initiate = TRUE` `initiate = FALSE` cases should be two different functions.

R/monitor.R
```diff
@@ -0,0 +1,151 @@
...
+    if (initiate) {
+        new_metrics <- vec_sort(df_metrics)
+    } else {
+        new_dates <- unique(df_metrics[[date_var]])
+        old_metrics <- pins::pin_read(board, metrics_pin_name)
+        old_metrics <- vec_slice(
+            old_metrics,
+            ! old_metrics[[date_var]] %in% new_dates
+        )
+        new_metrics <- vec_sort(vctrs::vec_rbind(old_metrics, df_metrics))
```
DavisVaughan, line 122: Am I reading this right that if you supply a new metric for the _same date_ as something that already exists in the pin, then it will overwrite the old one?

Maybe a note in the docs about this would be good.

It is also possible that some people won't want this behavior, or will want a way to opt out of it. I could see wanting to either:
- Error if I accidentally overlapped something in the pin
- Append it anyways

DavisVaughan, line 122: Do you really want to sort by the entire metrics data frame? Or just the date column? If the latter, you could use:

```r
new_metrics <- vctrs::vec_rbind(old_metrics, df_metrics)
new_metrics <- vec_slice(new_metrics, vctrs::vec_order(new_metrics[[date_var]]))
```

R/monitor.R
```diff
@@ -0,0 +1,151 @@
...
+    }
+
+    pins::pin_write(board, new_metrics, basename(metrics_pin_name))
```
DavisVaughan, line 125: What's up with `basename()`?

R/monitor.R
```diff
@@ -0,0 +1,151 @@
...
+
+}
+
+
+#' @rdname vetiver_compute_metrics
+#' @export
+vetiver_plot_metrics <- function(df_metrics,
+                                 date_var,
+                                 .estimate = .estimate,
+                                 .metric = .metric,
```
DavisVaughan, line 136:
```suggestion
                                 estimate = .estimate,
                                 metric = .metric,
```

DavisVaughan: Just because I think we only ever `.` arguments if they have the potential to conflict with named `...`

(I realize it is a little weird here because the column name is `.estimate`, so if you want to keep this then I wouldn't mind)

tests/testthat/test-monitor.R
```diff
@@ -0,0 +1,113 @@
+describe("vetiver_compute_metrics()", {
+
+    library(dplyr)
+    library(parsnip)
```
DavisVaughan, line 4: Jenny's talk on Monday would discourage the usage of `library()` anywhere inside the `testthat/` folder. Instead we'd just use `dplyr::` or `parsnip::`

Being inside the context set up by `describe()` means the objects like `lm_aug` are locally scoped, which is great, but the library calls will still affect other test files because they are global changes

DavisVaughan (commented):
Comments are mainly around two themes:

- Lots of yardstick specific tweaks to `vetiver_compute_metrics()`. I decided it was easier to dump in a rewrite of this function (you'll see that below), and I added comments about what exactly I was changing
- I think `vetiver_pin_metrics()` might be two different functions: `vetiver_initialize_metrics()` and `vetiver_update/append_metrics()`, so you can get rid of the special "first time" behavior you have to do with `initialize = TRUE`, which I think is a little hard to discover

R/monitor.R
```diff
@@ -0,0 +1,151 @@
...
+#' @param date_var The column in `data` containing dates or date-times for
+#' monitoring, to be aggregated with `unit`
+#' @param metric_set A [yardstick::metric_set()] function for computing metrics.
+#' Defaults to [yardstick::metrics()].
+#' @param df_metrics A tidy dataframe of metrics over time, such as created by
+#' `vetiver_compute_metrics()`.
+#' @param metrics_pin_name Pin name for where the *metrics* are stored (as
+#' opposed to where the model object is stored with [vetiver_pin_write()]).
+#' @param initiate Are you starting a new set of monitoring metrics to
+#' pin/store? Defaults to `FALSE`.
```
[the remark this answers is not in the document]
DavisVaughan: I am now somewhat convinced this could be two functions, and you could get rid of the `initialize` argument

R/monitor.R
juliasilge: It's because of some existing weirdness with pins on Connect, like rstudio/pins#613

R/monitor.R
juliasilge: I really wanted to inherit the docs for parameters from slider so I don't accidentally get out of sync. (This has happened in other packages.) Do you know of a way to inherit arg docs that have a different name, like using `.period` docs for `period`?

R/monitor.R
juliasilge: Turns out `pins::pin_write()` works perfectly well for the initiation process, so changed docs to use that and removed `initiate` entirely 👍

R/monitor.R
juliasilge: This definitely needed more explanation (and I did add an arg to opt out). Thanks! 👍

R/monitor.R
```diff
@@ -0,0 +1,151 @@
...
+#' library(ggplot2)
+#' vetiver_plot_metrics(new_metrics, date) +
+#'     scale_size(range = c(2, 4))
+#'
+#' @export
+vetiver_compute_metrics <- function(data,
+                                    date_var,
+                                    .period,
+                                    truth, estimate, ...,
+                                    metric_set = yardstick::metrics,
```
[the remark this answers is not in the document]
juliasilge: I checked with [he `R-CMD-check-hard` action like in broom and it was fine](https://github.com/tidymodels/broom/blob/main/.github/workflows/R-CMD-check-hard.yaml), so I'm going to risk it (vs. the NOTE).

juliasilge:
Thank you so much for the thorough review @DavisVaughan! Looks like this now:

``` r
library(vetiver)
library(dplyr)
#> 
#> Attaching package: 'dplyr'
#> The following objects are masked from 'package:stats':
#> 
#>     filter, lag
#> The following objects are masked from 'package:base':
#> 
#>     intersect, setdiff, setequal, union
library(parsnip)
data(Chicago, package = "modeldata")
Chicago <- Chicago %>% select(ridership, date, all_of(stations))
training_data <- Chicago %>% filter(date < "2009-01-01")
testing_data <- Chicago %>% filter(date >= "2009-01-01", date < "2011-01-01")
monitoring <- Chicago %>% filter(date >= "2011-01-01", date < "2012-12-31")
lm_fit <- linear_reg() %>% fit(ridership ~ ., data = training_data)

library(pins)
b <- board_temp()

## before starting monitoring, initiate the metrics and pin
## (for example, with the testing data):
original_metrics <-
    augment(lm_fit, new_data = testing_data) %>%
    vetiver_compute_metrics(date, "week", ridership, .pred, .every = 4L)
pin_write(b, original_metrics, "lm_fit_metrics")
#> Guessing `type = 'rds'`
#> Creating new version '20220523T032938Z-d51cc'
#> Writing to pin 'lm_fit_metrics'

## to continue monitoring with new data, compute metrics and update pin:
new_metrics <-
    augment(lm_fit, new_data = monitoring) %>%
    vetiver_compute_metrics(date, "week", ridership, .pred, .every = 4L) %>%
    vetiver_pin_metrics(b, "lm_fit_metrics")
#> Guessing `type = 'rds'`
#> Replacing version '20220523T032938Z-d51cc' with '20220523T032938Z-73a5d'
#> Writing to pin 'lm_fit_metrics'

library(ggplot2)
vetiver_plot_metrics(new_metrics) +
    scale_size(range = c(2, 4))
```



<sup>Created on 2022-05-22 by the [reprex package](https://reprex.tidyverse.org) (v2.0.1)</sup>

I have **one outstanding question** that I would still appreciate any ideas on. You [recommended taking out the `.` in the slider args](https://github.com/tidymodels/vetiver-r/pull/92#discussion_r871723755) but I did that to inherit for parameters from slider so the docs here don't accidentally get out of sync with slider or turn out wrong. (This has happened in other packages that I have observed.) 

- Do you know of a way to inherit arg docs that have a different name, like using .period docs for period?
- How would you balance the maintenance ease of inheriting the args vs. using kind of unnecessary dots in names?

R/monitor.R
DavisVaughan: It works AUTOMATICALLY for that specific case 🔥

R/monitor.R
DavisVaughan: Great!

R/monitor.R
DavisVaughan: Ah, I see. Feel free to mark as resolved and carry on then

R/monitor.R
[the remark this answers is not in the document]
juliasilge: I checked again (with CRAN back up this time) and it still looks OK to have `yardstick::metrics()` here. 🤞

PR #98, opened 2022-05-31
Add `vetiver_dashboard` template and function

juliasilge:
This PR adds a new dashboard for model monitoring. This includes an output format (`vetiver_dashboard()`) and an R Markdown template.


inst/rmarkdown/templates/vetiver_dashboard/skeleton/skeleton.Rmd
```diff
@@ -0,0 +1,91 @@
...
+  vetiver::vetiver_dashboard:
+    board: !expr pins::board_rsconnect()
+    name: 'julia.silge/hotel_rf'
+    storyboard: true
+    source_code: 'https://vetiver.rstudio.com'
+    display_pins: true
+params:
+    board: !r pins::board_rsconnect()
+    name: 'julia.silge/hotel_rf'
+    version: NULL
```
juliasilge, line 13: One of the big questions I have about this PR is how to avoid having all these things twice.

We need the `board` and `name` in the output format function because that is how the link for the pin is automatically generated, but we also need these in the `params` because they are used in the monitoring functions like `vetiver_pin_metrics()`.

inst/rmarkdown/templates/vetiver_dashboard/skeleton/skeleton.Rmd
```diff
@@ -0,0 +1,91 @@
...
+
+## load new validation data, for example from database or API
+validation_data_url <- 'https://tidymodels.org/start/case-study/hotels.csv'
+validation_df <- read_csv(validation_data_url) %>%
+    mutate_if(is.character, as.factor) %>%
+    arrange(arrival_date) %>%
+    filter(arrival_date > "2017-01-01") 
+
+v <- vetiver_pin_read(params$board, params$name, version = params$version)
+v_meta <- pin_meta(params$board, params$name)
```
juliasilge, line 32: This line is here because the next step will involve getting the metrics from training time out from the metadata.

tests/testthat/test-dashboard.R
```diff
@@ -0,0 +1,35 @@
...
+        edit = FALSE
+    ))
+
+    it("can create dashboard", {
+        expect_identical(rmd, path)
+        expect_true(file.exists(rmd))
+    })
+
+    it("can render dashboard", {
+        output_format <- vetiver_dashboard(pins::board_rsconnect(), "julia.silge/hotel_rf")
```
juliasilge, line 30: I'm not sure if I should use a real board in the tests like this -- probably not. See above about passing in these params.

juliasilge: This is the source of the failures right now.

juliasilge:
Closes #5

tests/testthat/test-dashboard.R
juliasilge: Fixed by @cderv in #103

juliasilge:
Closes #103 

Thank you so much @cderv! 🙏

juliasilge:
TODO:

- [ ] The pin URL doesn't seem to be working
- [ ] Make a better error if the pin doesn't exist and the user tries to knit the dashboard

inst/rmarkdown/templates/vetiver_dashboard/skeleton/skeleton.Rmd
```diff
@@ -0,0 +1,99 @@
...
+library(vetiver)
+library(pins)
+library(plotly)
+
+# packages needed for prediction
+library(parsnip)
+library(workflows)
+
+knitr::opts_chunk$set(echo = FALSE)
+pins <- get_vetiver_dashboard_pins()
```
cderv, line 25: Ok I think I have missed something important. 
```r
pins <- get_vetiver_dashboard_pins()
```
This will work well during knitting only and not interactively as the **knitr** option `knitr::opts_knit$get("vetiver_dashboard.pins")` won't be populated by the IDE - it will only be populated when the document is knitted and `vetiver::vetiver_dashboard()` is effectively evaluated. 

This may not be what you wanted considering that is was working as expected for you using `params`. Am I right ? 

Sorry for the oversight on this. The RStudio IDE in fact do a specific processing of the YAML `params` fields to make it available in the global environment for interactive execution of chunk. 

So we probably need to reconsider the current solution, aren't we ?

R/dashboard.R
```diff
@@ -0,0 +1,119 @@
...
+#' real model.
+#'
+#' @export
+vetiver_dashboard <- function(pins, display_pins = TRUE, ...) {
+
+    rlang::check_installed("flexdashboard")
+    dashboard_dots <- rlang::list2(...)
+
+    if (display_pins) {
+        v <- pin_read_version(pins$board, pins$name, pins$version)
```
cderv, line 34: This will indeed failed if the pins does not exist yet. (like when `pin_example_kc_housing_model()` has not be run prior to knit the document). 

If you want this format to run in example mode all the time, you coud try adding a `demo = TRUE` argument, and initialize the board in a temp board or local board before reading. But maybe that is too much...

R/dashboard.R
```diff
@@ -0,0 +1,119 @@
...
+        dashboard_dots <- vetiver_modify_navbar(dashboard_dots, v$metadata$url)
```
cderv, line 35: `v$metadata$url` I think this will be `NULL` when using a local board, won't it ? Maybe this should be checked and handle properly.

R/dashboard.R
```diff
@@ -0,0 +1,119 @@
...
+vetiver_modify_navbar <- function(dots, url) {
+    if(has_name(dots, "navbar")) {
+        navbar <- dots$navbar
+    } else {
+        navbar <- NULL
+    }
+    pin_navbar <- list(
+        title = "Model Pin",
+        icon = "fa-map-marker",
+        href = url,
```
cderv, line 79: With board local, `href = NULL`. this is what creates a wrong link in the resulting flexdashboard.

R/dashboard.R
```diff
@@ -0,0 +1,119 @@
...
+    navbar <- append(navbar, list(pin_navbar))
+    modifyList(dots, list(navbar = navbar))
+}
+
+
+#' @rdname vetiver_dashboard
+#' @export
+get_vetiver_dashboard_pins <- function() {
+    knitr::opts_knit$get("vetiver_dashboard.pins")
+}
```
cderv, line 91: See my comment above, this will return a non-NULL value only during knitting. In interactive mode (without rendering the all doc but running chunk by chunk), the knitr options won't be populated. 

I'll probably need to find another solution as you may not want this behavior.

cderv (commented):
I think I missed something important that we need to rediscuss. Please see my comment below.

R/dashboard.R
```diff
@@ -0,0 +1,134 @@
...
+dashboard_read_version <- function(board, name, version, call) {
+    if (!pins::pin_exists(board, name)) {
+        rlang::abort(c(
+            glue("The pin {glue::double_quote(name)} does not exist on your board"),
+            "*" = "Check the `pins` params in your dashboard YAML",
+            "i" = glue("To knit the example vetiver monitoring dashboard, \\
+                       execute `vetiver::pin_example_kc_housing_model()` to \\
+                       set up demo model and metrics pins")
+        ))
+    }
```
cderv, line 72: Nice way of handling the missing board! 

@juliasilge I am just wondering : You added a `call` argument, but you don't use it. Wasn't it to pass the `caller_env()` from `vetiver_dashboard()` into the call to `rlang::abort()` ? 

Maybe I misunderstood. Just sharing this feedback in case it was really the intention and it is just missing.

juliasilge: Thank you so much for the heads up @cderv! I got distracted on this and left it unfinished yesterday but have fixed in #108.

PR #100, opened 2022-06-06
Make a new arg `lockfile` for Dockerfile generation

### LINKED ISSUE #99: `vetiver_write_docker()` rewrites project level lockfile
As per title, if you have a project level renv.lock, `vetiver_write_docker()` rewrites it.

It would be interesting to have the same behavior as the package [dockerfiler](https://github.com/ThinkR-open/dockerfiler), that writes a renv.lock.dock for the Dockerfile.

I can implement it myself if you also believe it makes sense...

juliasilge:
Closes #99 

@ggpinto Thank you so much for the feedback! 🙌  Would you like to take a look at this and see if it would fit your use case?

``` r
library(vetiver)
library(pins)
b <- board_temp(versioned = TRUE)
vetiver_pin_write(b, v)
#> Creating new version '20220606T221019Z-9ca24'
#> 
vetiver_write_plumber(b, "cars_linear")

vetiver_write_docker(v)
cat(readr::read_lines("Dockerfile"), sep = "\n")
#> # Generated by the vetiver package; edit with care
#> 
#> FROM rocker/r-ver:4.2.0
#> ENV RENV_CONFIG_REPOS_OVERRIDE https://packagemanager.rstudio.com/cran/latest
#> 
#> RUN apt-get update -qq && apt-get install -y --no-install-recommends \
#>   git \
#>   libcurl4-openssl-dev \
#>   libgit2-dev \
#>   libicu-dev \
#>   libsodium-dev \
#>   libssl-dev \
#>   make
#> 
#> COPY vetiver_renv.lock renv.lock
#> RUN Rscript -e "install.packages('renv')"
#> RUN Rscript -e "renv::restore()"
#> COPY plumber.R /opt/ml/plumber.R
#> 
#> ENTRYPOINT ["R", "-e", "pr <- plumber::plumb('/opt/ml/plumber.R'); pr$run(host = '0.0.0.0', port = 8000)"]
```

<sup>Created on 2022-06-06 by the [reprex package](https://reprex.tidyverse.org) (v2.0.1)</sup>

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,7 @@
 # vetiver (development version)
 
+* The lockfile created by `vetiver_write_docker()` can now be named via the argument `lockfile`, and its default is `vetiver_renv.lock` (#100).
+
 # vetiver 0.1.5
 
 * Add functions for model monitoring (#92).
--- a/R/write-docker.R
+++ b/R/write-docker.R
@@ -4,14 +4,16 @@ DEFAULT_RSPM <-  "https://packagemanager.rstudio.com"
 #' Write a Dockerfile for a vetiver model
 #'
 #' After creating a Plumber file with [vetiver_write_plumber()], use
-#' `vetiver_write_docker()` to create a Dockerfile plus an `renv.lock` file
-#' for a pinned [vetiver_model()].
+#' `vetiver_write_docker()` to create a Dockerfile plus a `vetiver_renv.lock`
+#' file for a pinned [vetiver_model()].
 #'
 #' @inheritParams vetiver_api
 #' @param plumber_file A path for your Plumber file, created via
 #' [vetiver_write_plumber()]. Defaults to `plumber.R` in the working directory.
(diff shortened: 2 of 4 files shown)


ggpinto:
@juliasilge that was fast! Yes, that would work perfectly!

I see now another problem: it should have a line EXPOSE 8000, otherwise the container would exit automatically, should I open another issue?

juliasilge:
@ggpinto That has not been my experience with the containers I have built. I believe that `EXPOSE` is just for documenting?

- https://stackoverflow.com/questions/22111060/what-is-the-difference-between-expose-and-publish-in-docker
- https://www.ipspace.net/kb/DockerSvc/10-Exposed.html

Do you find the documenting useful? Or are you using this in a different way?

ggpinto:
@juliasilge  I am not the most experienced with containers, when I try to create a container using Docker Desktop, if there is no EXPOSE in the Dockerfile I don't see any port options

image

PR #113, opened 2022-06-24
New function to create RStudio Connect bundle

### LINKED ISSUE #88: Create `vetiver_create_rsconnect_bundle()` or similar
We already have [`vetiver_deploy_rsconnect()`](https://vetiver.tidymodels.org/reference/vetiver_deploy_rsconnect.html) but let's make it easier for folks to use git-backed or RSC-API-backed deployment by creating a bundle ready for that.

- https://docs.rstudio.com/connect/user/structuring-content/#content-bundles
- An example: <https://github.com/sol-eng/bike_predict/tree/main/API>

juliasilge:
Closes #88 

This PR adds a new function to create an RStudio Connect bundle. This is an alternative to the `vetiver_deploy_rsconnect()` function, and will work best with the (not on CRAN) connectapi R package, or else just curl. A difference from the `vetiver_deploy_rsconnect()` function is that it does not actually deploy, but only gets ready for deployment. This is because I did not want to duplicate the functionality of connectapi.


blairj09 (commented):
Sorry I was late on this! But this all checks out to me!

PR #115, opened 2022-06-30
Add model packaging/unpackaging capability to support more model types

bjfletcher:
# Background

My data science team use H2O models in some of our work, using the R library from H2O.

When we've trained a model using H2O, the R object is actually a reference to the trained model which resides in the filesystem or a separate runtime with a REST interface.

If we want to store (pin) the model to the S3 bucket (board), we need a mechanism to read the model from filesystem into a R object.

If we want to retrieve the model from the S3 bucket, we also need a mechanism to write the model from R object to the filesystem, and get a reference to it.

# Proposal

This PR introduces a couple of new functions `model_package` and `model_unpackage` that are called by `vetiver_pin_write` and `vetiver_pin_read` respectively. These are optional through S3 methods.

For example, when we want to store the trained model by running `vetiver_pin_write`, it needs to read the model from filesystem into a R object and then write (pin) it to the S3 bucket (board). This is achieved with this S3 method:

```r
#' @rdname model_package
#' @export
model_package.h2o <- function(model) {
    f <- tempfile()
    h2o::h2o.saveModel(model, path = dirname(f), filename = basename(f))
    file_size <- file.info(f)$size
    fd <- file(f, 'rb', raw = T)
    raw_model <- readBin(fd, what = 'raw', n = file_size, endian = 'little')
    close(fd)
    file.remove(f)
    structure(raw_model, class = class(model))
}
```

When we want to retrieve the model by running `vetiver_pin_read`, it needs to write the model (in R object) into the filesystem, load it to H2O and then get the reference to the model residing in the filesystem.

```r
#' @rdname model_unpackage
#' @export
model_unpackage.h2o <- function(model) {
    model <- structure(model, class = NULL) # get the original object
    f <- tempfile()
    fd <- file(f, 'wb')
    writeBin(model, con = fd, endian = 'little')
    close(fd)
    loaded_model <- h2o::h2o.loadModel(f)
    file.remove(f)
    loaded_model
}
```

(You may wonder "hang on, we have `vetiver_prepare_model`!. This is run at `vetiver_model` creation time, for use locally as well as for writing to the board. This is different to serialising / deserialising the model to / from the board.)

A couple of unit tests are included, with a question mark of whether there is a better way to unregister S3 methods (I've [asked on StackOverflow](https://stackoverflow.com/questions/72820860/how-to-unregister-a-s3-method)). But replacing them with the default methods seems ok to me.

Thanks :)


simonpcouch:
Thanks for the PR!

This kind of functionality is definitely on our roadmap. For now, our plan is to incorporate methods like these (for H2O, but also other models) into a standalone package, that packages like vetiver and [parsnip](https://github.com/tidymodels/parsnip) could then incorporate under the hood (or users could call themselves if they prefer). The current draft of a description we've been passing around is:

> R holds most objects in memory. However, some models store their data in locations that are not included when one uses `save()` or `saveRDS()`.  [package_name] provides a common API to capture this information, situate it within a portable model `bundle`, and restore it for use in new settings.

While our current work on the package is quite experimental, please let us know if we can tag you once the package is at a stage where we're ready for feedback—would love to hear opinions on the interface from folks like you. I'd anticipate the timeline on that being 2-4 months out.

juliasilge:
@bjfletcher We have put together a new package that handles tasks like this called [bundle](https://github.com/rstudio/bundle). Would you like to take a look and see if this will fit your use case? After bundle goes to CRAN, we'll update vetiver to use `bundle()` in the `vetiver_prepare_model` methods.

juliasilge:
Let us know if you get a chance to look at bundle @bjfletcher! 🙌

PR #122, opened 2022-07-08
Handle `type` in `vetiver_pin_metrics()`

### LINKED ISSUE #118: Specify pin type in vetiver_pin_metrics?
Currently vetiver_pin_model does not have an options to specify pin type. I think this means that pins will be stored as Rds files. However, in pins-python the type argument is required. Should we add a type argument to vetiver_pin_model? Alternatively, it could re-use the type of the pin it is writing a new version to.

`pin_write` can also take other arguments. For example, with the rstudio board a person might do `pin_write(..., access_type="all")`. I wonder if we should pass `...` args in to `pin_write`?

(I think a similar situation arises with not being able to pass  args like access_type occurs with `vetiver_pin_write`)

See https://pins.rstudio.com/reference/board_rsconnect.html#public-pins

juliasilge:
Closes #118 

This PR makes two main changes (for the better):

- `vetiver_pin_metrics()` now finds the type of the original pin and updates with the same `type`
- The dots are now passed through to `vetiver_pin_metrics()` so you can do things like `access_type = "all"`

I had to reorganize how these three functions are documented because of using different dots in two of them. It's probably better this way anyway.


R/monitor.R
```diff
@@ -123,17 +87,89 @@ vetiver_compute_metrics <- function(data,
...
                                 .index = .index,
-                                overwrite = FALSE) {
+                                overwrite = FALSE,
+                                ...) {
     .index <- enquo(.index)
     .index <- eval_select_one(.index, df_metrics, "date_var")
     new_dates <- unique(df_metrics[[.index]])
 
+    old_type <- pins::pin_meta(board, metrics_pin_name)$type
+    if (! old_type %in% c("arrow", "rds")) {
```
machow, line 168: What's the reason CSV is not supported? Is it because `pin_read()'ing` a CSV won't read in date columns as dates?

(I think a similar challenge occurred on the python side, but I had it attempt to coerce the date index to a date for better or worse 😬, so might be good to align behavior one way or the other..!)

R/monitor.R
```diff
@@ -152,7 +188,13 @@ vetiver_pin_metrics <- function(board,
         vctrs::vec_order(new_metrics[[.index]])
     )
 
-    pins::pin_write(board, new_metrics, basename(metrics_pin_name))
+    pins::pin_write(
+        board,
+        new_metrics,
+        basename(metrics_pin_name),
+        type = old_type,
```
machow, line 195: vetiver-python has it use the `old_type` only if the user doesn't specify e.g. `type="arrow"`. I wonder if that could be useful, if users want to switch the type?

machow (changes requested):
I left a couple comments about differences between this and the current vetiver-py PR implementation. Less concerned with which approach is right, than surfacing the differences in case it's useful to sync them up!

R/monitor.R
juliasilge: I outlined this a bit over in #118 but in R, it's not only the date but also the metrics columns and such (pins uses the base R `read.csv()` which is famously troublesome). There would be quite a bit of coercing and type fixing to do with CSV, so much so that I think enforcing something like arrow is a good move. I am open to users bringing this up later again, though, if CSV is important for some situations.

machow: That makes sense! I noticed the presence of the infamous strings are being read as factors situation while checking pin_read w/ CSV 😅

R/monitor.R
juliasilge: Sure, that sounds good. Added in 66af701b81f595c94835d3b7b91363e45ad7e4ad

R/monitor.R
juliasilge: I've added #123 to track interest in CSV metrics storage for the R side 👍

PR #140, opened 2022-08-31
Move plumber to suggests

### LINKED ISSUE #136: plumber in Suggests?
I understand if this is a route you don't want to go down, but there are situations where deploying *without* plumber makes sense, such as on AWS Lambda (see #128). It would be nice to avoid installing extra things in that case, and {plumber} comes with a number of dependencies, including {sodium}, which has a `libsodium` system requirement (and thus adds another step to the process).

Would you consider moving {plumber} to Suggests for {vetiver}, and only adding it to the `vetiver_renv.lock` when it makes sense to do so (ie, for almost every deployment path other than Lambda, probably)?

It probably isn't a HUGE issue, but I can imagine getting to a point where we want to shave every little thing out of that container that we can.

juliasilge:
Closes #136


juliasilge:
Hmmm, I am realizing that the [testing strategy for plumber](https://github.com/rstudio/vetiver-r/blob/main/tests/testthat/helper-plumber.R) might not work without plumber in Imports.

juliasilge:
I'm going to close this for now, and we can continue to collect feedback on this change for prioritization.

juliasilge:
@jonthegeek I realized that I could move the plumber testing helper into a regular R file in the package, which made this doable after all. Do you want to take a look at this and see if you notice anything that would be a problem for the Lambda use case?

jonthegeek (commented):
Oh wow this is great! I'll be back into the Lambda deployment work tomorrow or Monday, I think, so I'll try it out more formally then, but I don't see anything that should be a problem!

juliasilge:
I'm going to merge it in now then, and we can revisit any problems that come up.

PR #145, opened 2022-09-26
Use bundle for deployment

juliasilge:
This PR adds support for [bundle](https://github.com/rstudio/bundle) to vetiver, for:

- tidymodels, plus stacks
- xgboost
- caret


DESCRIPTION
```diff
@@ -48,6 +46,8 @@ Suggests:
     flexdashboard,
     ggplot2,
     httpuv,
+    httr,
+    jsonlite,
```
juliasilge, line 50: Run up against the 20 packages in Imports again 😞 so I moved these to Suggests and them use them conditionally. These are mostly used only when making a _prediction_ from R so not a terrible option to move to Suggests.

tests/testthat/test-xgboost.R
```diff
@@ -21,15 +21,13 @@ test_that("can pin an xgboost model", {
     b <- board_temp()
     vetiver_pin_write(b, v)
     pinned <- pin_read(b, "cars2")
-    ## NOT EQUAL because of serialization issues
```
simonpcouch, line 24: Awesome to see this test issue resolved. :)

simonpcouch (approved):
This looks great to me!

PR #154, opened 2022-10-28
Use renv to check packages at `vetiver_pin_read()`

### LINKED ISSUE #4: Use renv to track needed packages and versions and store a hash
In #2 I talk about _loading_ packages but we need to figure out how to have platforms like RSC know what packages to _install_. RSC currently uses a manifest, not renv, so we'll need to figure out some details here.

If a user has a file that looks like this:

```r
library(deploytidymodels)
library(pins)
library(plumber)

model_board <- board_rsconnect()

#* @plumber
function(pr) {
    pr %>%
        pr_model(model_board, "julia.silge/biv_svm", type = "class")
}
```

we currently will not successfully install the right packages for prediction into the app.

@topepo has already put work into creating an `renv.lock` file from a model to track what packages/versions are needed for prediction; I think this can either be used or repurposed.

juliasilge:
Closes #4 

I think what renv returns (like JSON) is small enough that we can store the whole thing, not a hash. This will help us give more useful information to the user.

### DIFF
--- a/R/pin-read-write.R
+++ b/R/pin-read-write.R
@@ -33,11 +33,24 @@
 #'
 #' @export
 vetiver_pin_write <- function(board, vetiver_model, ...) {
+
+    pkgs <- vetiver_required_pkgs(vetiver_model$metadata$required_pkgs)
+    lockfile_pkgs <-
+        renv::snapshot(
+            lockfile = NULL,
+            packages = pkgs,
+            prompt = FALSE,
+            force = TRUE
+        )
+
     pins::pin_write(
         board = board,
-        x = list(model = vetiver_model$model,
-                 ptype = vetiver_model$ptype,
-                 required_pkgs = vetiver_model$metadata$required_pkgs),
+        x = list(
+            model = vetiver_model$model,
+            ptype = vetiver_model$ptype,
+            required_pkgs = vetiver_model$metadata$required_pkgs,
+            renv_pkgs = lockfile_pkgs
+        ),
         name = vetiver_model$model_name,
         type = "rds",
         description = vetiver_model$description,
@@ -62,7 +75,15 @@ vetiver_pin_read <- function(board, name, version = NULL) {
     pinned <- pins::pin_read(board = board, name = name, version = version)
     meta   <- pins::pin_meta(board = board, name = name, version = version)
 
-    ## TODO: add subset of renv hash checking
+    lockfile_pkgs <-
+        renv::snapshot(
+            lockfile = NULL,
+            packages = vetiver_required_pkgs(pinned$required_pkgs),
+            prompt = FALSE,
+            force = TRUE
+        )
+
+    ## TODO: compare pinned$renv_pkgs to new lockfile_pkgs
 
     new_vetiver_model(
         model = pinned$model,
--- a/R/write-docker.R
+++ b/R/write-docker.R
@@ -58,8 +58,7 @@ vetiver_write_docker <- function(vetiver_model,
         ""
     )
 
-    pkgs <- unique(c(docker_pkgs, vetiver_model$metadata$required_pkgs))
-    pkgs <- setdiff(pkgs, drop_pkgs)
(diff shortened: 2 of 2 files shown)


tests/testthat/test-pin-read-write.R
```diff
@@ -106,3 +109,33 @@ test_that("can read a versioned model with metadata", {
...
+            lockfile = renv$renv_lockfile_init(project = NULL)
+        ),
+        name = v$model_name,
+        type = "rds",
+        description = v$description,
+        metadata = v$metadata$user,
+        versioned = v$versioned
+    )
+
+    expect_message(vetiver_pin_read(b, "cars5", check_renv = TRUE))
```
juliasilge, line 140: How can I test that this message is being created? If I am reading the renv code right, I believe [it is sent to stdout (or maybe stderr?)](https://github.com/rstudio/renv/blob/37f4e97e0cff7dfddfcef9d2de702f7c45e056fb/R/pretty.R#L133). The message looks about like:

```
The following package(s) do not match your model:

# CRAN ===============================
- R6          [2.5.1 -> *]
- Rcpp        [1.0.9 -> *]
- cli         [3.4.1 -> *]
- etc etc etc
```
I'd like to write a test like `expect_message(regexp = "do not match your model")` but apparently this isn't a message. When I tried `expect_output()` it failed, so apparently this isn't output at all. What am I missing?

R/pin-read-write.R
```diff
@@ -30,14 +33,34 @@
...
+#' vetiver_pin_write(model_board, v, check_renv = TRUE)
 #'
 #' @export
-vetiver_pin_write <- function(board, vetiver_model, ...) {
+vetiver_pin_write <- function(board, vetiver_model, ..., check_renv = FALSE) {
+
+    lockfile <- character(0)
+
+    if (check_renv) {
+        pkgs <- vetiver_required_pkgs(vetiver_model$metadata$required_pkgs)
```
juliasilge, line 46: TODO: This should only be the packages for the model itself, not infrastructure like plumber and such.

juliasilge:
From discussions with @isabelizimm, we should choose names for the args (currently `check_renv` and `lockfile`) that are less renv-specific, because this kind of information is likely also useful for the Python metadata.

R/pin-read-write.R
juliasilge: We decided to try doing the model packages plus vetiver only, and see how that goes.

R/pin-read-write.R
```diff
@@ -57,12 +82,32 @@ vetiver_pin_write <- function(board, vetiver_model, ...) {
...
+        } else {
+            cli::cli_warn(c(
+                "There is no lockfile stored with {.val {name}}:",
+                "i" = "Use {.arg check_renv = TRUE} when you save your model to your board"
+            ))
+        }
+    }
+
+    meta$user <- list_modify(meta$user, renv_lock = zap())
+    if (is_empty(meta$user)) names(meta$user) <- NULL
```
juliasilge, line 110: This is needed because `list_modify()` returns an empty _named_ list here, not just `list()`.

R/pin-read-write.R
```diff
@@ -30,9 +33,31 @@
...
+#' vetiver_pin_write(model_board, v, check_renv = TRUE)
 #'
 #' @export
-vetiver_pin_write <- function(board, vetiver_model, ...) {
+vetiver_pin_write <- function(board, vetiver_model, ..., check_renv = FALSE) {
+
+    renv_lock <- NULL
+
+    if (check_renv) {
+        pkgs <- c(vetiver_model$metadata$required_pkgs, "vetiver")
```
juliasilge, line 46: This does _not_ include all the packages needed for the API like plumber, pins, etc but I think that is the right decision. That makes these the packages needed for the _model only_ (not necessarily the deployment).

juliasilge:
Thanks for taking a look at this PR @kevinushey! Let me know if any more context would be helpful.

kevinushey (approved):
LGTM!

Issue #127, opened 2022-07-25
Challenges when sharing models due to package versions


tyluRp:
I am currently sharing development models using the example workflow:

```r
library(vetiver)
library(pins)

# connect to model board
b <- board_s3(
  bucket = "models", 
  prefix = "label/", 
  region = structure("us-east-1", tags = list(type = "scalar"))
)

# checkout the version you want
v <- vetiver_pin_read(
  board = b, 
  name = "label", 
  version = "20220701T182136Z-9905d"
)

# create some dummy data
new_data <- data.frame(
  predictor = c(
    "some text",
    "more text"
  )
)

# occasionally, models require packages be loaded
load_pkgs(v$metadata$required_pkgs)
#> Registered S3 method overwritten by 'tune':
#>   method                   from   
#>   required_pkgs.model_spec parsnip

# make predictions
predict(v, new_data)
#> # A tibble: 2 × 1
#>   .pred_class
#>   <fct>      
#> 1 TRUE       
#> 2 FALSE
```

However, we have found that this doesn't always work due to package versions being different from computer to computer. Occasionally, a user will call `predict` on new data and an error is returned as a result of package version differences. We have come to this conclusion because some computers can run the exact same code without error.

Is there a suggested workflow or any guidance on how to share development models? We are thinking of making a function like `with_model_env` that would leverage the `renv.lock` to run the model with the same set of packages temporarily but wanted to see if there is a better approach.

tyluRp:
Understood, that makes sense. In the few times where package versions have been an issue, we have been able to resolve by installing the latest of all required packages using:

```r
install.packages(v$metadata$required_pkgs)
```

machow:
Just a high-level thought, it seems like there may be three levels requirements that could be stashed with the model:

* **current**: the name of the package (no version number)
* **package-style requirements**: the name of the package pinned to version number (or something like that)
* **deployment-style requirements**: a lockfile pinning the exact version of all installed packages

juliasilge:
We've got new support for renv now implemented via #154! If you would like to try it out, you can install with `devtools::install_github("rstudio/vetiver-r")` and then use the new `check_renv = TRUE` argument when you write or read. You can see a model that was [stored using this new argument here](https://colorado.posit.co/rsc/content/3d10a0b5-4712-4a07-aa53-a2045ee1998d). Click on "Raw metadata" to see the new metadata like `required_pkgs` and `renv_lock`.

Let me know if you run into further issues or questions with this.

PR #157, opened 2022-11-04
Bundle and load renv at startup

juliasilge:
Related to #154, #4, and #127

Once I get this right and merged in, I'll work on extending how we use renv (not only writing Dockerfiles, but also when writing/reading the model to storage).


R/zzz.R
```diff
@@ -0,0 +1,14 @@
...
+
+.onLoad <- function(libname, pkgname) {
+    load_renv_env()
+    renv$renv_platform_init()
+    renv$renv_envvars_init()
+    renv$renv_log_init()
+    renv$renv_methods_init()
+    renv$renv_filebacked_init()
+    renv$renv_libpaths_init()
+    renv$renv_patch_init()
```
juliasilge, line 11: Does this seem right? Do I need all of it? If I did not include this, I would errors like this when trying to use functions from the renv environment:

```
Error in get(scope, envir = `_renv_filebacked`) : 
object 'settings' not found
```

R/renv.R
```diff
@@ -0,0 +1,6 @@
+renv <- new.env()
```
juliasilge, line 1: For my own learning at some point, this did _not_ work as `rlang::new_environment()`. If I tried to use that here, I would errors like this when building the package, i.e. it couldn't load the code into the environment:

```
Error in renv_abi_check <- function(packages = NULL, ..., libpaths = NULL,  : 
  could not find function "<-"
```

juliasilge:
In the `inst/resources/renv.R` file that is generated/embedded, I had to do a find-and-replace to change all the instances of `tail()` to `utils::tail()`. Maybe it makes more sense to use `@importFrom utils tail` instead, although I think that will require one of my PRECIOUS slots for imports going to utils.

R/renv.R
hadley: I'd suggest putting this in `zzz.R` to be closer to where you actually use it.

R/zzz.R
juliasilge: Why doesn't packrat need this?

juliasilge:
Clearly this is not working yet:

```
  Loading this package had a fatal error status code 1
  Loading log:
  Error: package or namespace load failed for ‘vetiver’:
   .onLoad failed in loadNamespace() for 'vetiver', details:
    call: NULL
    error: "" does not exist
```

I can build and check the package locally in the current state.

R/zzz.R
juliasilge: I took all this out, because it was resulting in the package not building. There are other problems now, but this seems like it was a bad idea.

juliasilge:
I took out the stuff in `.onLoad` that was causing the problems for loading the package. 👍 

However, now I am having the problems with actually running any code from the `renv` environment. Trying to call `renv$snapshot` results in this traceback:
```
Error in get(scope, envir = `_renv_filebacked`) : 
object 'settings' not found
14. get(scope, envir = `_renv_filebacked`)
13. renv_filebacked_envir(scope)
12. renv_filebacked_get(scope, path)
11. filebacked("settings", path, renv_settings_read_impl)
10. renv_settings_read(path)
9. renv_settings_get(project, name)
8. settings$r.version(project = project)
7. settings$r.version(project = project) %||% getRversion()
6. renv_lockfile_init_r_version(project)
5. renv_lockfile_init_r(project)
4. renv_lockfile_init(project)
3. renv_lockfile_create(project, libpaths, type, packages)
2. renv$snapshot(project = path, lockfile = lockfile, packages = pkgs, 
prompt = FALSE, force = TRUE) at write-docker.R#63
1. vetiver_write_docker(v, tmp_plumber, tempdir())
```

@kevinushey do you have any ideas on this? 

- How does packrat handle these kinds of settings? I looked but couldn't figure it out.
- After taking a look, does this (vendoring the renv code) strike you as the best option, vs. new exported functionality in renv itself?

kevinushey:
Ah, I think this is a small bug in `renv` in terms of being embeddable. Normally, some stuff needs to happen in its `.onLoad()` hook; however, that's not getting called here. I think I can fix this in `renv`; let me take a look.

kevinushey:
I made a change in `renv` that seems to fix the issue for me locally:

https://github.com/rstudio/renv/commit/f672288ff7a261928d69d454aa483dd280fb788b

Can you let me know if you have better luck with 0.16.0-23?

kevinushey:
> Maybe it makes more sense to use @importFrom utils tail instead, although I think that will require one of my PRECIOUS slots for imports going to utils.

`renv` uses quite a few things from `utils`:

https://github.com/rstudio/renv/blob/f672288ff7a261928d69d454aa483dd280fb788b/NAMESPACE#L44-L72

I should probably avoid doing this to help the embedding scenario, or perhaps provide a helper script that ensures `renv` can always "see" `utils`.

kevinushey:
I've added some tooling to the development version of `renv` to make this easier. Now, you should be able to just run:

```
renv:::vendor()
```

inside the package directory, and `renv` will set up the necessary infrastructure for you. For example:

```
kevin@MBP-P2MQ:~/r/pkg/vetiver-r [main]
$ R -s -e "renv:::vendor()"
# Cloning renv 0.16.0-26 from https://github.com/rstudio/renv ...Done!
#
# A vendored copy of `renv` was created at: "~/r/pkg/vetiver-r/inst/vendor/renv.R"
# The `renv` auto-loader was generated at:  "~/r/pkg/vetiver-r/R/renv.R"
#
# Please add `renv$initialize()` to your package's `.onLoad()`
# to ensure that `renv` is initialized on package load.
#
```

juliasilge:
I think something is still up with loading the package in this way @kevinushey. 😕 I am getting these errors (the same as when I tried to put things like `renv$renv_platform_init()` in `.onLoad`):

```
  Loading this package had a fatal error status code 1
  Loading log:
  Error: package or namespace load failed for ‘vetiver’:
   .onLoad failed in loadNamespace() for 'vetiver', details:
    call: NULL
    error: "" does not exist
```

I only see these errors on CI, not locally. It reminds me of [this kind of problem](https://github.com/tidyverse/glue/pull/254) that we saw in glue and stringr, where one package is being loaded while another namespace is "sort of" loaded. Maybe related?

kevinushey:
Doh; I can reproduce locally if I run the tests if `renv` is not installed, so there's likely something in `renv` code that expects to be able to find itself as an installed package. I should be able to fix this; sorry for the trouble!

kevinushey:
Sorry for the false starts -- the development version on `renv` should properly handle this now. You should be able to just run:

```
remotes::install_github("rstudio/renv")
renv:::vendor()
```

and be ready to go; if you bump into another issue please let me know!

kevinushey:
Sorry for the false starts! I think this is finally ready to go; I've added some proper tests to `renv` as well and tests are passing there, so I'm fairly confident this should work in `vetiver` now. You can update to the latest with:

```
remotes::install_github("rstudio/renv")
renv:::vendor()
```

kevinushey:
Progress! Although now it looks like the failures are related to the repository URLs. This might be related to some internal code that `renv` tries to use when running tests, which tries to make sure it always runs without any CRAN repositories visible. I think I need to turn this off for the embedded mode.

kevinushey:
Should be good to go! Indeed, there shouldn't be any licensing issues.

The only thing I'd add is that if you plan to use internal renv functions, please add a few unit tests in case I change something in a backwards-incompatible way (although I'll do my best not to!)

juliasilge:
> if you plan to use internal renv functions, please add a few unit tests in case I change something in a backwards-incompatible way (although I'll do my best not to!)

Definitely will! The current use in Dockerfiles is pretty well tested via snapshots, etc, and I'll be sure to test additional use. Just to make this explicit here, that means *I* will find out if a change in renv breaks vetiver functionality, not *you*, which is backwards compared to the normal reverse dependency checks for R packages. 👍

PR #160, opened 2022-11-14
Fix how `vetiver_ptype()` works for models with interactions

### LINKED ISSUE #158: vetiver_ptype gets congfused if lm has interaction terms
I kept getting the following error when trying to make vetiver_model of a lm object, which I narrowed down to be an issue of ptype

> > vetiver_ptype(model)
> Error in `[.data.frame`(model$model, pred_names) : 
>   undefined columns selected

If I read the codebase correctly, it might be an issue of not taking out interaction terms of lm models before building the ptype resulting in the error. I'm not confident enough in the elegance of my solution to make a PR out of this, but have provided a solution that does seem to eliminate the error.

Minimal working example (with a possible way to fix the issue demonstrated)
```
library(vetiver)

data(mtcars)
model <- lm(mpg ~ cyl*gear+disp, data=mtcars)

# this does not work
vetiver_ptype(model)

# removing interaction terms from pred_names
vetiver_ptype_edited.lm <- function(model, ...) {
  pred_names <- attr(model$terms, "term.labels")
  # taking away interaction terms, they will always have a ':' in them
  pred_names_without_interaction <- pred_names[!grepl(':',pred_names)]
  ptype <- vctrs::vec_ptype(model$model[pred_names_without_interaction])
  tibble::as_tibble(ptype)
}

# this works
vetiver_ptype_edited.lm(model)
```

juliasilge:
Closes #158 

This uses the approach that hardhat took in tidymodels/hardhat#175 to identify when a model has any interactions:

``` r
library(vetiver)

cars_lm <- lm(mpg ~ cyl*gear + disp, data = mtcars)
vetiver_ptype(cars_lm)
#> # A tibble: 0 × 3
#> # … with 3 variables: cyl <dbl>, gear <dbl>, disp <dbl>
```

<sup>Created on 2022-11-14 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -2,6 +2,8 @@
 
 * Now vendor renv directly in package (#157).
 
+* Fixed how `vetiver_ptype()` finds predictors for models (`lm()` and `glm()`) with interactions (#160).
+
 # vetiver 0.1.8
 
 * Trailing slashes are now removed from `vetiver_endpoint()` (#134).
--- a/R/lm.R
+++ b/R/lm.R
@@ -13,7 +13,7 @@ vetiver_prepare_model.lm <- function(model) {
 #' @rdname vetiver_create_ptype
 #' @export
 vetiver_ptype.lm <- function(model, ...) {
-    pred_names <- attr(model$terms, "term.labels")
+    pred_names <- preds_lm_ish(model)
     ptype <- vctrs::vec_ptype(model$model[pred_names])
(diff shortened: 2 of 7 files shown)


juliasilge:
This currently works for `lm()` and `glm()` but I think this approach will work for other models in the future, if needed.

PR #161, opened 2022-11-17
Change `vetiver_write_docker()` to write plumber and Dockerfile in one step

juliasilge:
We have been discussing the usefulness of adding a new function that writes both the plumber file and Dockerfile (along with the lockfile) in one step, rather than always [requiring folks to use a two-step process](https://vetiver.rstudio.com/learn-more/deploy-with-docker.html#create-api-and-docker-artifacts) even when their needs are straightforward:

> I think a wrapper function would be great. I was actually going to suggest that `required_pkgs(b)` should be the default value for `additional_pkgs`, but then realised that the board is not passed to `vetiver_write_docker`. A wrapper function would sort this out.

_Originally posted by @miguellacerda in https://github.com/rstudio/vetiver-r/issues/159#issuecomment-1311595700_


juliasilge:
This new function works like so:

``` r
library(vetiver)
library(pins)
b <- board_temp()
vetiver_pin_write(b, v)
#> Creating new version '20221117T220709Z-ebc45'
#> 

vetiver_write_docker_proj(b, "cars_linear")
#> - R             [* -> 4.2.1]
#> 
```

<sup>Created on 2022-11-17 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>

It creates a plumber file, Dockerfile, and lockfile at `path` (defaults to working directory). @miguellacerda are you able to try this out and see what you think?

The big question in my mind is what to name a function like this. Some possibilities:
- `vetiver_write_docker_proj()` (or `_project`) but it's not really a project in any way we use that word otherwise?
- `vetiver_write_docker_requirements()` (or `_reqs`) but that seems confusing with system requirements
- `vetiver_setup_docker()` but it's really just writing files
- what else?

It would be not ideal to rename the current `vetiver_write_docker()` but maybe that needs a more specific name (`vetiver_write_dockerfile()`???) and then this function can use `vetiver_write_docker()`. Handling a name change like this is not fun but could be doable.

isabelizimm:
I don't mind docker project! Some other ideas off the top of my head, although I don't really love any of them off the bat:

- `vetiver_write_docker_bundle` (although maybe confusing with R `bundle`) 
- `vetiver_write_docker_assets`
- `vetiver_write_docker_components`

Secondary option is to make writing a plumber file an argument inside `vetiver_write_docker`?

juliasilge:
Another possibility: this is actually what `vetiver_write_docker()` should do, and we don't need a separate function but to change what the existing function does.

miguellacerda:
@juliasilge I tried it out and it works beautifully! Much slicker than before!

In terms of names, I like @isabelizimm's suggestion of adding an option to write the plumber file as a boolean argument within the existing `vetiver_write_docker` function.  But now that I think about it, why are the write functions prefixed with "vetiver"? Is there a reason that you didn't go with `write.pin` and `write.docker` like many other write functions in R?

juliasilge:
It's not really possible to add writing the plumber file as an arg, because you need the board and name for the plumber file. However, we could change `vetiver_write_docker()` to take the board and name as arguments instead. This would be a **breaking change** but one that gets us some significant interface improvements. The arguments for writing a plumber file are all still exposed in almost the exact same way as before.

@miguellacerda want to take the new changes here for a spin and see what you think?

juliasilge:
Proposed new interface:

``` r
library(vetiver)
library(pins)
b <- board_temp()
vetiver_pin_write(b, v)
#> Creating new version '20221122T235556Z-ebc45'
#> 
vetiver_write_docker(b, "cars_linear")
#> - R             [* -> 4.2.1]
#> 
```

<sup>Created on 2022-11-22 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>

Error if folks try the old interface:

``` r
library(vetiver)
library(pins)
b <- board_temp()
vetiver_pin_write(b, v)
#> Creating new version '20221122T235649Z-ebc45'
#> 
vetiver_write_plumber(b, "cars_linear")
vetiver_write_docker(v)
#> Error:
#> ! The `vetiver_model` argument of `vetiver_write_docker()` was
#>   deprecated in vetiver 0.2.0 and is now defunct.
#> ✖ This function no longer uses your model object directly.
#> ℹ Instead, use the `board` and `name` where your model is stored.

#> Backtrace:
#>     ▆
#>  1. └─vetiver::vetiver_write_docker(v)
#>  2.   └─lifecycle::deprecate_stop(...) at vetiver-r/R/write-docker.R:70:8
#>  3.     └─lifecycle:::deprecate_stop0(msg)
#>  4.       └─rlang::cnd_signal(...)
```

<sup>Created on 2022-11-22 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>

jrosell:
Breaking changes should be avoided.
If it can write more things why name it with "docker"? 

What about:
- vetiver_write() 
- vetiver_snapshot()
- or another two word function name

miguellacerda:
> @miguellacerda want to take the new changes here for a spin and see what you think?

Works perfectly! I think the improved interface is definitely worth it.

isabelizimm:
I think this is worth the breaking change! Having a plumber file is a strong requirement for the Dockerfile, so it makes sense to be able to generate these 2 files at once. It seems to make deploying a Dockerfile through vetiver clearer.

On naming: since the end goal of this function is to generate all the assets to build and run a Dockerfile, I do think that having Docker in the name is the right choice. One suggestion is to change `vetiver_write_docker` -> `vetiver_write_dockerfile`, which I don't mind, although it's a little long. But, I'm also happy as-is with `vetiver_write_docker`.

jrosell:
> I think this is worth the breaking change! Having a plumber file is a strong requirement for the Dockerfile, so it makes sense to be able to generate these 2 files at once. It seems to make deploying a Dockerfile through vetiver clearer.
> 
> On naming: since the end goal of this function is to generate all the assets to build and run a Dockerfile, I do think that having Docker in the name is the right choice. One suggestion is to change `vetiver_write_docker` -> `vetiver_write_dockerfile`, which I don't mind, although it's a little long. But, I'm also happy as-is with `vetiver_write_docker`.

Then I vote for vetiver_setup_docker. The reason is that maybe in the future we can have other vetiver_setup_*

Issue #55, opened 2021-10-29
Consider options for more flexible `ptype` specification


juliasilge:
The `ptype` concept here in vetiver is used as an **input data prototype**, for checking that the format of new data at prediction conforms to what the model had at training time.

It is already possible to totally turn off this checking, when appropriate to a certain use case.

In discussions with @t-kalinowski he explained to me how it will likely be important for keras, etc to be able to have more flexible, _partially_ defined input data shapes/prototypes such as an array that is fixed in some dimensions but could be any length in another dimension. I don't currently have great ideas for this, something flexible between "no checking" and "very strict checking".

juliasilge:
We may be able to [use `vctrs::new_partial()` here](https://vctrs.r-lib.org/reference/new_partial.html).

juliasilge:
I have been working on initial support for keras in #164 and here is where I plan to go initially:

- Error if `length(model$inputs) > 1` to exclude multi-input keras models for now. The handling of `body` in the POST request is going to get pretty complex for non-rectangular data, so let's wait to see how folks use it and if they want support for the multi-input model.
- Require folks to _either_ set `save_prototype = FALSE` for no checking of new data or input some `prototype_data` (like for ranger), so like `vetiver_model(keras_fit, "cars-keras", prototype_data = data.frame(x_train)[1,])`. This still is "rectangular data" only. Let's see how important the higher dimensional case (like images or other non-rectangular data) is for folks and understand how they want this to behave.

I am happy to hear 👂 about what the most important use cases for folks are, so we can build out to more complex support of what keras is capable of.

PR #164, opened 2022-11-21
Add support for keras models

### LINKED ISSUE #144: Add support for keras
Now that [bundle](https://rstudio.github.io/bundle/) is on CRAN, we can add support for keras.

juliasilge:
Addresses #144


R/keras.R
```diff
@@ -0,0 +1,42 @@
...
+#' @rdname vetiver_create_meta
+#' @export
+vetiver_create_meta.keras.engine.training.Model <- function(model, metadata) {
+    vetiver_meta(metadata, required_pkgs = "keras")
+}
+
+#' @rdname vetiver_create_ptype
+#' @export
+vetiver_ptype.keras.engine.training.Model <- function(model, ...) {
+    NULL
```
juliasilge, line 17: @t-kalinowski one question I have is whether it would be better to have a default of `NULL` for the input data prototype (schema) which is better for non-rectangular data like images or whether we should have a default of rectangular data and get some data from the model or user. Either way, users can opt out of the default via the `save_prototype` argument, so this is a question about defaults.

R/keras.R
```diff
@@ -0,0 +1,42 @@
...
+    attach_pkgs(vetiver_model$metadata$required_pkgs)
+}
+
+#' @rdname handler_startup
+#' @export
+handler_predict.keras.engine.training.Model <- function(vetiver_model, ...) {
+
+    function(req) {
+        new_data <- vetiver_type_convert(req$body, vetiver_model$ptype)
+        new_data <- as.matrix(new_data)
```
juliasilge, line 38: @t-kalinowski do you have a suggestion for better default handling of the new data coming into the API? This seems to work for the options I've tried out.

tests/testthat/helper-redact.R
```diff
@@ -17,3 +17,20 @@ redact_vetiver <- function(snapshot) {
 redact_port <- function(snapshot) {
     snapshot <- gsub(port, "<port>", snapshot, fixed = TRUE)
 }
+
+
+clean_python_tmp_dir <- function(variables) {
```
juliasilge, line 22: @t-kalinowski Thank you so much for your suggestions on this!

t-kalinowski: Glad it worked!

t-kalinowski (approved):

juliasilge:
Thank you again for this review @t-kalinowski! I outlined some of where we are and where we might head from here in #55. To sum up:

- This now errors on creating the `vetiver_model()` object for multi-input keras models, because the current strategy for the POST endpoint will not handle that well. Let's hear from users about how important this is.
- The default behavior is to require the user to pass in some `prototype_data` (this is similar to ranger and all of the Python implementation) _or_ to explicitly turn off checking of new data with `save_prototype = FALSE`. If someone wants to build a model with non-rectangular data like images, this is currently what they will have to do. Again, let's hear from users about how important this is and what they want to do.

juliasilge:
I have made some good headway on how to deploy this to Connect today. Some notes:

- The automatic dependency discovery for Python didn't work for me at all, I think because I am on an M1 and all my TensorFlow stuff is specific for ARM architecture. I could solve this by deploying the plumber file together with a minimal `requirements.txt`:

```
keras
tensorflow
```

I did [successfully deploy a model](https://colorado.posit.co/rsc/connect/#/apps/54eb9375-8a2e-4864-982f-c27928c9368d/)! 🎉 

- Sadly, the model can't predict. 😞 From the logs, I suspect that the problem is that I trained and serialized the model on my M1. For the next step, I am going to try to train the model on Colorado Workbench and then deploy from there. I'm currently having trouble install TensorFlow on Workbench, however.

juliasilge:
Still TODO: create a minimal `requirements.txt` to deploy with the `plumber.R` for Connect.

juliasilge:
This is working now on Connect, I believe!
https://colorado.posit.co/rsc/content/9c37e89e-c094-41b8-813c-2f636dd98e51

I will work on a Docker container in a different PR.

PR #165, opened 2022-11-30
New function `vetiver_setup_docker()`

juliasilge:
Closes #161 

This PR tries out the option of keeping both `vetiver_write_plumber()` and `vetiver_write_docker()` in the package, and creating a new function `vetiver_setup_docker()`.

### What I like about this:

- No breaking change
- The [getting started](https://vetiver.rstudio.com/get-started/deploy.html) page could be edited to not include generating a plumber/FastAPI file at all, or to de-emphasize it (how/why to edit these files could go in a more advanced article) 
- We have some nice communication/marketing going on with `vetiver_deploy_rsconnect()` (deployment in a one liner) vs. `vetiver_setup_docker()` (this says we have set it up for you but you still have some work to do)

### What I still am not sure if like:

- Is "setup" a verb or even a word at all??? I think `vetiver_set_up_docker()` is even worse, though.


juliasilge:
@jrosell It turns out you are full of good ideas! I think we are going to try out `vetiver_prepare_docker()`.

juliasilge:
Here is what this looks like now:

``` r
library(vetiver)
library(pins)
b <- board_temp()
vetiver_pin_write(b, v)
#> Creating new version '20221201T182403Z-ebc45'
#> 
vetiver_prepare_docker(b, "cars_linear")
#> - R             [* -> 4.2.1]
#> 
```

<sup>Created on 2022-12-01 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>

This writes a plumber file, an renv lockfile, and a Dockerfile.

juliasilge:
@isabelizimm do you want to take a look at this and see if this is actually the way we want to go?

PR #166, opened 2022-12-07
Change `ptype` to `prototype`

### LINKED ISSUE #84: For discussion: `ptype` naming convention could be confusing to python users
I have had some exposure to both the Python and R vetiver packages. I want to start a discussion around the `ptype` parameter name (e.g. `save_ptype`, `ptype_data`, etc.). 

I think it could be confusing for python users. In pandas, there is the notion of [dtypes](https://pandas.pydata.org/docs/reference/api/pandas.DataFrame.dtypes.html). 

At first glance, I thought ptype was some how related. This is quickly cleared up if you read the docs... However, I think there could be some initial confusion for those who do not read the docs first.

Could it be reasonable to change the name to be more verbose? E.g. `save_prototype` and `prototype_data`? I think this would make it more clear that it is not related to types, and is instead related to prototypes. Going even one step further, is prototype the right word? When I think of prototypes I think of early / pre-release versions of a product.

juliasilge:
Closes #84 

This PR deprecates the `save_ptype` argument:

``` r
library(vetiver)
v <- vetiver_model(cars_lm, "cars-linear", save_ptype = FALSE)
#> Warning: The `save_ptype` argument of `vetiver_model()` is deprecated as of vetiver
#> 1.0.0.
#> ℹ Please use the `save_prototype` argument instead.
v <- vetiver_model(cars_lm, "cars-linear", save_prototype = FALSE)
```

<sup>Created on 2022-12-07 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>

The argument name change is propagated into the internal functions, but I have not currently changed the names [of these functions](https://rstudio.github.io/vetiver-r/reference/vetiver_create_ptype.html), used almost entirely by someone writing code to support a new model in vetiver.


juliasilge:
@isabelizimm do you want to take a look at this and see what you think with respect to similar changes in Python?

R/vetiver-model.R
```diff
@@ -47,13 +48,22 @@ vetiver_model <- function(model,
...
+            "vetiver_model(save_ptype)",
+            "vetiver_model(save_prototype)"
+        )
+        save_prototype <- save_ptype
+    }
     if (is_null(description)) {
         description <- vetiver_create_description(model)
     }
-    ptype <- vetiver_create_ptype(model, save_ptype, ...)
+    ptype <- vetiver_create_ptype(model, save_prototype, ...)
```
isabelizimm, line 66: Do we still want to call the attribute `ptype`?

isabelizimm:
One more comment: as I put together [#](https://github.com/rstudio/vetiver-python/pull/138), should `check_ptype` be updated to `check_prototype` on the vetiver API, since it is not an internal function?

R/vetiver-model.R
juliasilge: My concern about changing that is we would need to set up some handling to read models pinned with the old name. If we keep the name the same as it is now, we can continue to read models pinned with old versions of vetiver. What do you think?

juliasilge:
> One more comment: as I put together https://github.com/rstudio/vetiver-python/pull/138, should check_ptype be updated to check_prototype on the vetiver API, since it is not an internal function?

Since it's user-facing, yes. Thanks for the catch!

R/vetiver-model.R
isabelizimm: It feels a bit strange that while the argument is called `save_prototype`, it doesn't save `prototype`, it saves `ptype`. I don't think it's too much of a mental leap, but worth noting. Overall, I don't think it is as big of a concern as the backwards compatibility/maintainability, so I'm good with it staying ptype 👍

juliasilge: OK, it's actually way simpler to look for either `ptype` or `prototype` than I was imagining; see 9d2076ad2c2bd70f939907da3591a63678390bff.

With this installed I can read old pinned models just fine:

``` r
library(vetiver)
library(pins)

b <- board_rsconnect()
#> Connecting to RSC 2022.11.0 at <https://colorado.rstudio.com/rsc>
v <- b %>% vetiver_pin_read("julia.silge/wow-another-mtcars-model")
v
#> 
#> ── julia.silge/wow-another-mtcars-model ─ <butchered_lm> model for deployment 
#> An OLS linear regression model using 10 features
```

<sup>Created on 2022-12-09 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>

I'm going to change what the internal element is called. 👍

juliasilge: I ended up changing more than I initially thought I would, but I believe this is fairly future-proof and won't cause unexpected problems.

juliasilge:
@isabelizimm do you mind doing another review of this, and see if we are happy with this level of changes?

isabelizimm (approved):
This looks good to me at a high level!

PR #171, opened 2023-01-18
minor edit to keras example

t-kalinowski:
avoid implicit promotion of `y` from a 1d array to 2d array


juliasilge (approved):
Thank you! 🙌

PR #176, opened 2023-02-17
Moved `required_pkgs` to plain text metadata

### LINKED ISSUE #168: Refactor metadata, specifically move `required_pkgs`
We've decided to keep as much metadata as possible outside of the binary bundle, stored in plain text (YAML). As of today, here is what is wrapped up into the binary object:

https://github.com/rstudio/vetiver-r/blob/8875142d10aff1d55dc16170725f114a629b2884/R/pin-read-write.R#L38-L40

In R, we don't have any tool like [pydantic](https://docs.pydantic.dev/) that can completely/accurately store that can store the signature/schema/ptype of a dataframe in plain text like JSON or YAML. This means that for now, we will keep the `prototype` / `ptype` in the `list()` that gets serialized to binary when storing.

It is possible/easier to move `required_pkgs` to the `user` metadata slot, since it's just plain text.

juliasilge:
Closes #168 

I can still successfully read "old" models:

``` r
library(vetiver)
library(pins)
b <- board_connect()
#> Connecting to Posit Connect 2023.01.0 at <https://colorado.posit.co/rsc>
v <- b %>% vetiver_pin_read("julia.silge/scooby-doo")
v$metadata     
#> $user
#> list()
#> 
#> $version
#> [1] "62570"
#> 
#> $url
#> [1] "https://colorado.posit.co/rsc/content/4300d6fb-f761-4b12-ab81-bd85baada06e/_rev62570/"
#> 
#> $required_pkgs
#> [1] "LiblineaR" "parsnip"   "recipes"   "workflows"
```

<sup>Created on 2023-02-16 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>


juliasilge:
You can see how this looks in the metadata for [this example model](https://colorado.posit.co/rsc/connect/#/apps/3d10a0b5-4712-4a07-aa53-a2045ee1998d/access) (click through on "Raw metadata").

isabelizimm (commented):
This all seems like a good way to move `required_pkgs` into plaintext. TIL `zap()` ⚡ should there be an "old pins" test, or not worth it to test across older versions?

juliasilge:
Thank you @isabelizimm; that is probably a good idea to include. I added an additional test in 4043a02d75c362d8983feb92a5c6c43360fe7203.

PR #186, opened 2023-03-10
Add Sagemaker deployment support

DyfanJones:
This PR allows vetiver to deploy to AWS SageMaker

Ticket: https://github.com/rstudio/vetiver-r/issues/175


DyfanJones:
@juliasilge Let me know when you have updated finished aligning the functions to vetiver api :) I will add in some unit tests. Are you happy with some mock tests? I was thinking to help prevent having to call AWS and that.

juliasilge:
I was working on this today and I ran into a problem around how versioning is done on GHCR.

This works fine:

```
FROM rocker/r-ver:4.1.0
```

But this does not [because `4.1.0` is not a tag there](https://github.com/rocker-org/rocker-versioned2/pkgs/container/r-ver/versions?filters%5Bversion_type%5D=tagged):

```
## does not work
FROM ghcr.io/rocker-org/r-ver:4.1.0
```

Instead I would need to do this:

```
FROM ghcr.io/rocker-org/r-ver:4.1
```

juliasilge:
Looks like there [isn't GHCR support back that far](https://github.com/rocker-org/rocker-versioned2/wiki) maybe?

juliasilge:
This is looking pretty good! There is still some work to be done around the tags for the model and endpoint.

R/sagemaker.R
```diff
@@ -0,0 +1,481 @@
...
+#' @param compute_type The [CodeBuild](https://aws.amazon.com/codebuild/)
+#' compute type as a character. Defaults to `BUILD_GENERAL1_SMALL`.
+#' @param role The IAM role name for CodeBuild to use as a character. Defaults
+#' to the SageMaker Studio execution role.
+#' @param bucket The S3 bucket to use for sending data to CodeBuild as a
+#' character. Defaults to the SageMaker SDK default bucket.
+#' @param vpc_id ID of the VPC that will host the CodeBuild project
+#' (such as `"vpc-05c09f91d48831c8c"`).
+#' @param subnet_ids List of subnet IDs for the CodeBuild project
+#' (such as `list("subnet-0b31f1863e9d31a67")`).
```
juliasilge, line 123: Is this right @DyfanJones? i.e. folks should use this argument like `list("subnet-0b31f1863e9d31a67")`?

R/sagemaker.R
```diff
@@ -0,0 +1,481 @@
...
+#' @param security_group_ids List of security group IDs for the CodeBuild
+#' project (such as `list("sg-0ce4ec0d0414d2ddc")`).
```
juliasilge, line 125: Is this one also right?

R/sagemaker.R
```diff
@@ -0,0 +1,481 @@
...
+#' Currently only Amazon EBS gp2 storage volumes are supported.
+#' @param model_data_download_timeout The timeout value, in seconds, to download
+#' and extract model data from Amazon S3.
+#' @param wait A logical for whether to wait for the endpoint to be deployed.
+#' Defaults to `TRUE`.
+#' @rdname vetiver_sm_build
+#' @export
+vetiver_sm_endpoint <- function(model_name,
+                                endpoint_name = NULL,
+                                instance_type = "ml.t2.medium",
```
juliasilge, line 299: I added a non-`NULL` default value here, because the instance type is not optional. This filters up to the main `vetiver_deploy_sagemaker()` function as well. I think we have two options here:

- set a default like this (thoughts on whether this is the right one?)
- require the user to put in a value (no default) in which case it should go with the other required arguments in `vetiver_deploy_sagemaker()`

The second option would make the function signature look something like this, which does strike me as weird:

```r
vetiver_deploy_sagemaker <- function(board,
                                     name,
                                     instance_type,
                                     version = NULL,
                                     predict_args = list(),
                                     docker_args = list(),
                                     repo_name = glue("vetiver-sagemaker-{name}"),
                                     compute_type = c(
                                         "BUILD_GENERAL1_SMALL",
                                         "BUILD_GENERAL1_MEDIUM",
                                         "BUILD_GENERAL1_LARGE",
                                         "BUILD_GENERAL1_2XLARGE"
                                     ),
                                     ...) {
<<CODE>>
}
```

I don't like how the pin arguments are getting mixed up with others. What do you think @DyfanJones?

R/sagemaker.R
DyfanJones: Yes that is correct, it ultimately goes to this formatting:

```r
list(
      "vpcId" = list("vpc1"),
      "subnets" = list("subnet_ids1"),
      "securityGroupIds" = list("security_group_ids1")
)
```

Which mimics this behaviour: https://github.com/aws-samples/sagemaker-studio-image-build-cli/blob/master/sagemaker_studio_image_build/cli.py#L58-L67

Would it be worth changing vetiver's parameter to vpc_config to align with the other SageMaker functions so it is consistent  🤔

DyfanJones: If we did take the `vpc_config` parameter route it would looks something like:

```r
vetiver_sm_build <- function(board,
                             name,
                             version = NULL,
                             path = fs::dir_create(tempdir(), "vetiver"),
                             predict_args = list(),
                             docker_args = list(),
                             repository = NULL,
                             compute_type = c(
                                 "BUILD_GENERAL1_SMALL",
                                 "BUILD_GENERAL1_MEDIUM",
                                 "BUILD_GENERAL1_LARGE",
                                 "BUILD_GENERAL1_2XLARGE"
                             ),
                             role = NULL,
                             bucket = NULL,
                             vpc_config = list(),
                             log = TRUE,
                             ...) {
    check_installed("smdocker")
    compute_type <- arg_match(compute_type)

    # create dockerfile using
    # https://github.com/rocker-org/rocker-versioned2/pkgs/container/r-ver
    docker_args <- list_modify(docker_args, port = 8080)
    predict_args <- list_modify(predict_args, path = "/invocations")

    vetiver_prepare_docker(
        board = board,
        name = name,
        version = version,
        path = path,
        predict_args = predict_args,
        docker_args = docker_args
    )

    image_uri <- smdocker::sm_build(
        repository = repository,
        compute_type = compute_type,
        role = role,
        dir = path,
        bucket = bucket,
        vpc_id = vpc_config$vpcId,
        subnet_ids = vpc_config$Subnets %||% list(),
        security_group_ids = vpc_config$SecurityGroupIds %||% list(),
        log = log,
        ...
    )

    return(image_uri)
}
```

R/sagemaker.R
DyfanJones: Sorry I am not a 100% on what the 2 options are sorry :s

From what I understood:

- Is having `instance_type` with a default"ml.t2.medium" ok?
    - I am not sure, I don't think `ml.t2.medium` isn't available in all aws regions. From checking https://aws.amazon.com/sagemaker/pricing/ it looks like it is not available in regions: af-south-1, ap-east-1, ap-east-1, ap-southeast-3, ap-northeast-3, eu-south-1, eu-north-1, me-south-1, me-central-1 and us-gov-east-1. I believe that is the reason why sagemaker sdk doesn't specify the `instance_type`: https://github.com/aws/sagemaker-python-sdk/blob/master/src/sagemaker/estimator.py#L1319-L1428
- What should we do around the `pins` and aws parameters mix in the function `vetiver_deploy_sagemaker`.
    - We could do something similar to the `vetiver_prepare_docker` function and have dedicated listing parameters. Maybe something like:

```r
vetiver_deploy_sagemaker <- function(board,
                                     name,
                                     instance_type,
                                     version = NULL,
                                     predict_args = list(),
                                     docker_args = list(),
                                     build_args = list(),
                                     ...) {
<<CODE>>
}
```

R/sagemaker.R
```diff
@@ -0,0 +1,481 @@
...
+        version = version,
+        predict_args = predict_args,
+        docker_args = docker_args,
+        repository = repo_name,
+        compute_type = compute_type,
+        bucket = board$bucket
+    )
+
+    args <- list2(...)
+    tags <- list_modify(args$tags, "vetiver")
```
DyfanJones, line 84: > This is looking pretty good! There is still some work to be done around the tags for the model and endpoint.

I believe the issues is down to this that. It should be in the following format

```r
list(
    list(
        'Key'='string',
        'Value'='string'
     )
)
```
This is my bad for not documenting it correctly, sorry about that

DyfanJones: Got an initial idea for making tagging easier, I will push it up so i can do an initial test on my sagemaker environment.

juliasilge:
Carl Boettiger [says the rocker images should be exempt from Docker rate limits now](https://fosstodon.org/@cboettig@ecoevo.social/110022868241804516).

R/sagemaker.R
juliasilge: I think we should keep it the same as what `smdocker::sm_build()` does, whatever that is. Do you think it would be better to change both the smdocker and vetiver args to be more like the other SageMaker functions?

R/sagemaker.R
juliasilge: OK, sounds like we should _not_ specify an instance type for this, which means [we _should_ put it up with the required arguments](https://design.tidyverse.org/args-data-details.html#args-data-details). It feels real weird to split up the pin arguments and put `instance_type` before `version`, though. Would it be weird/awkward to have it first? Or we could put `version` into a new `build_args()` argument because `vetiver_sm_build()` uses it. That does divide it up but isn't as jarring in the function reference.

R/sagemaker.R
DyfanJones: Happy to keep it as it is. `smdocker` currently mimics it's python counter part, if there was a strong desire to move to vpc_config I can look into adding it to the `smdocker` :)

R/sagemaker.R
DyfanJones: Make sense for it to be a required argument :) I think either approach would be acceptable. Probably need to consider it's api in comparison to `vetiver_deploy_rsconnect `. Should help with future deploy methods for different platforms when having that consistent-ish api template across the `vetiver_deploy_*` functions :)

juliasilge:
This is now working for me in SageMaker, including tags! 🎉 Would you be up for trying this out yet again @DyfanJones? I have [updated this demo for how this looks as of today](https://github.com/juliasilge/vetiverdemo/blob/main/biv-svm/train-sm-docker.Rmd).

I think all that is still to be decided is how to set up the function signature for `vetiver_deploy_sagemaker()`.

DyfanJones:
> This is now working for me in SageMaker, including tags! 🎉 Would you be up for trying this out yet again @DyfanJones? I have [updated this demo for how this looks as of today](https://github.com/juliasilge/vetiverdemo/blob/main/biv-svm/train-sm-docker.Rmd).
> 
> I think all that is still to be decided is how to set up the function signature for `vetiver_deploy_sagemaker()`.

It is working for me :) What are your thoughts on the missing is the ModelDataUrl from the Sagemaker Model? Should we leave it for now and if needed add it later?

Final note, currently `vetiver_sm_delete` will remove: endpoint config, endpoint and model. Should there parameters that allows users to only drop core stages of vetiver:

```r
vetiver_sm_delete <- function(object, delete_model = TRUE, delete_endpoint = TRUE) {
    check_installed(c("smdocker", "paws.machine.learning"))

    config <- smdocker::smdocker_config()
    sagemaker_client <- paws.machine.learning::sagemaker(config)

    endpoint_name <- object$model_endpoint

    if (!is.null(delete_endpoint)) {
        tryCatch(
            {
                endpoint_config_name <- sagemaker_client$describe_endpoint(
                    EndpointName = endpoint_name
                )$EndpointConfigName
                sagemaker_client$delete_endpoint_config(endpoint_config_name)
            },
            error = function(err) {
                cli::cli_warn("Unable to delete {.val {endpoint_name}} endpoint configuration.")
            }
        )
        sagemaker_client$delete_endpoint(endpoint_name)
    }
    if (!is.null(delete_model)) {
        sagemaker_client$delete_model(endpoint_name)
    }
    return(invisible(TRUE))
```
So it reflects the `vetiver_sm_endpoint`, `vetiver_sm_model` functions?

juliasilge:
I updated the `vetiver_sm_delete()` function like you suggested. 👍 

> What are your thoughts on the missing is the ModelDataUrl from the Sagemaker Model?
It would be nice if we can get it in now, while we're thinking about it. I am having a hard time finding where that model screen even is, though. 🙈 

Other than that, I think it's time to start thinking about tests. I'll start on some of these to show how I am planning to do mocking, and then it would be great for you to take a look too.

DyfanJones:
Sounds good to me

DyfanJones:
The model page is on the inference section, then model, you should see the model you have created, by selecting it you should get metadata around the model.

Just need to think of how to ge the s3 uri to the model. Will have a look tomorrow, if you know how please go ahead and implement :)

juliasilge:
Do you think there might be something wrong with how the milliseconds are formatted in `base_name_from_image()`? One time I saw this error:

> 1 validation error detected: Value 'vetiver-sagemaker-starstudded-felin-2023-03-16-20-55-21-679e+09' at 'modelName' failed to satisfy constraint: Member must satisfy regular expression pattern: ^[a-zA-Z0-9]([\-a-zA-Z0-9]*[a-zA-Z0-9])?

It seems to mostly be working fine, though.

DyfanJones:
Hmm interesting, never hought that would happen, will have to see if there is a better way to get milliseconds

juliasilge:
I think that's a good option for providing a "ModelDataUrl"!

I've gotten a start on testing, and I think it's going pretty well. Do you want to take a look and see if there are any other tests you think are important?

DyfanJones:
They look good, I think we should have a test for `base_name_from_image` so that it is returning the correct format (that the issue you raised before :) )

juliasilge:
## Question 1

Was this approach for the "ModelDataUrl" working for you? When I try to run it, I get tags like this:

```
[[1]]
[[1]]$Key
[1] "my-tag"

[[1]]$Value
[1] "support-vector-machine"


[[2]]
[[2]]$Key
[1] "vetiver:pin_board"

[[2]]$Value
s3://sagemaker-vetiver-anthophobic-longhornbeetle/


[[3]]
[[3]]$Key
[1] "vetiver:r-ver"

[[3]]$Value
[1] ‘4.1.0’
```

And then I get this error from SageMaker:

> Error: ValidationException (HTTP 400). Model data URL "s3://sagemaker-vetiver-anthophobic-longhornbeetle/" is invalid.

## Question 2

Take a look at the test I wrote for `base_name_from_image()`; does that capture the edge case that caused my one-time problem before? I am still fuzzy on what was happening.

DyfanJones:
Ah looking at ModelDataUrl again, it looks like we can't use it (that is my bad for not reading the documentation probably) as we don't use tar.gz we can just leave it in the tags.

> ModelDataUrl (string) --
The S3 path where the model artifacts, which result from model training, are stored. This path must point to a single gzip compressed tar archive (.tar.gz suffix). The S3 path is required for SageMaker built-in algorithms, but not if you use your own algorithms. For more information on built-in algorithms, see [Common Parameters](https://docs.aws.amazon.com/sagemaker/latest/dg/sagemaker-algo-docker-registry-paths.html).

Sorry I am not near my laptop to revert this change.

>Take a look at the test I wrote for base_name_from_image(); does that capture the edge case that caused my one-time problem before? I am still fuzzy on what was happening.

- We could take 2 approaches, the first is the one you have done which is fine :) or we could just check the pattern using regex so something like.

```r
test_that("can get base name from image", {
    image_uri <- "999999999999.dkr.ecr.us-east-2.amazonaws.com/vetiver-sagemaker-example-model:2023-03-17"

    expect_true(grepl(
        "vetiver-sagemaker-example-model-[0-9]{4,}-[0-9]{2,}-[0-9]{2,}-[0-9]{2,}-[0-9]{2,}-[0-9]{2,}-[0-9]{3,}"
        base_name_from_image(image_uri)
    ))
})
```

I think we just want to make sure it doesn't do this:

`vetiver-sagemaker-starstudded-felin-2023-03-16-20-55-21-679e+09`

juliasilge:
I think this is looking GREAT and may be about ready to merge! 🎉 

@DyfanJones when you get a chance, would you install one more time and try this out yourself again? I am [using code like this](https://github.com/juliasilge/vetiverdemo/blob/main/biv-svm/train-sm-docker.Rmd).

I am going to share this with folks internally who work with SageMaker but I'd like this merged with the docs published before I share it around more.

DyfanJones:
Perfect, I will be able try it out tomorrow 😄 . I agreed it is looking pretty good, really happy in how easy it allows users to push to sagemaker.

DyfanJones:
Noticed an issue with `vetiver_sm_delete`. Currently it will delete both endpoint and model even if delete_model or delete_endpoint are set to `FALSE`. Added some unit tests to check if delete_model or delete endpoint have been called when parameter are set to `FALSE`.

DESCRIPTION
```diff
@@ -68,6 +70,7 @@ Suggests:
     rpart,
     rsconnect,
     slider (>= 0.2.2),
+    smdocker (>= 0.1.1.9000),
```
juliasilge, line 73:
```suggestion
    smdocker (>= 0.1.2),
```

tests/testthat/test-sagemaker.R
```diff
@@ -0,0 +1,204 @@
...
+    expect_snapshot(
+        out <- sm_create_endpoint(client, endpoint_name, config_name, wait = FALSE)
+    )
+
+    expect_equal(out, endpoint_name)
+})
+
+test_that("can delete SageMaker endpoints", {
+    mock_delete_endpoint_config <- mockery::mock(abort("No config!"), list())
+    mock_delete_model <- mockery::mock()
```
juliasilge, line 112: Can we put something here in the mocked function? I find it kind of confusing to parse when we haven't set any values to return.

tests/testthat/test-sagemaker.R
```diff
@@ -0,0 +1,204 @@
...
+    mockery::stub(vetiver_sm_delete, "sagemaker_client$delete_endpoint_config", mock_delete_endpoint_config)
+    mockery::stub(vetiver_sm_delete, "sagemaker_client$delete_endpoint", list())
+    mockery::stub(vetiver_sm_delete, "sagemaker_client$delete_model", mock_delete_model)
+
+    object <- vetiver_endpoint_sagemaker("vetiver-sagemaker-example-model")
+
+    expect_snapshot(vetiver_sm_delete(object))
+    expect_true(vetiver_sm_delete(object))
+    # Check if delete model and endpoint have been called
+    expect_equal(environment(mock_delete_endpoint_config)$call_no, 2)
```
juliasilge, line 123: Can we use `mockery::expect_called()` here and below instead of manually getting the env?

juliasilge: You may already be familiar with this, but this vignette outlines that:
https://cran.rstudio.com/web/packages/mockery/vignettes/mocks-and-testthat.html

DyfanJones: Will change it now

tests/testthat/test-sagemaker.R
DyfanJones: Updated it to have the original value within the mock function :)

juliasilge (approved):
Thank you SO MUCH for all this work @DyfanJones! 🙌 I am super excited for this functionality to be available for SageMaker users.

PR #205, opened 2023-05-18
Update pkgdown navbar

juliasilge:
We had discussion this week that it would be good to have an additional link to https://vetiver.studio.com on the navbar. I also think it would be good to add a more prominent link to the Python docs.

@isabelizimm here is what I ended up doing! I think "Learn more" is probably a better option that something about "main site" (although I know we say that a lot internally). I bet you can do [this kind of navbar with Quarto](https://quarto.org/docs/websites/website-navigation.html), right?

### DIFF
--- a/_pkgdown.yml
+++ b/_pkgdown.yml
@@ -2,14 +2,15 @@ template:
   bootstrap: 5
   package: tidytemplate
 
+
 navbar:
+  structure:
+    left:  [intro, reference, news, vetiver]
   components:
-    home: ~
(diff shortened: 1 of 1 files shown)


isabelizimm:
I can do the same dropdown menu with links to R and vetiver.rstudio.com on the Python side. I like just using the name of the site, rather than trying to rename it as something else. 👍

juliasilge:
I messed up my branches 🙈 and these changes got merged in with #206 FYI

PR #210, opened 2023-05-24
Verify that `plumber_file` exists before creating Dockerfile

JosiahParry:
Currently there is no check to see if the `plumber_file` argument leads to a real plumber file. It is possible to create a Dockerfile with a non-existent plumber file at present. See reprex below. 

This PR adds a check to verify that the plumber file does indeed exist. 

``` r
library(vetiver)

dir.create("inst/vetiver", recursive = TRUE)
vetiver_write_docker(v, path = "inst/vetiver")
#> - R             [* -> 4.3.0]
#> 

list.files("inst/vetiver")
#> [1] "Dockerfile"
list.files()
#> [1] "inst" "vetiver_renv.lock"
```

### DIFF
--- a/R/write-docker.R
+++ b/R/write-docker.R
@@ -64,6 +64,16 @@ vetiver_write_docker <- function(vetiver_model,
                                  additional_pkgs = character(0)) {
 
     ellipsis::check_dots_empty()
+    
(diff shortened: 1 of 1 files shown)


juliasilge (approved):
Thank you so much @JosiahParry! 🙌

PR #213, opened 2023-06-01
Add Docker support for keras

### LINKED ISSUE #144: Add support for keras

juliasilge:
Closes #144


juliasilge:
As I wrote out in #144 it's going to be pretty challenging to handle this on different architectures (ARM vs AMD) in a very automated way. For now, let's improve the situation via documentation and some warnings of choices we know _won't_ work and then look for problems folks run into moving forward.

PR #221, opened 2023-06-09
Changing `board_rsconnect()` to `board_connect()` as the former was deprecated in pins 1.1.0.

aminadibi:
Addresses https://github.com/rstudio/vetiver-r/issues/188 and related to https://github.com/rstudio/pins-r/issues/678


juliasilge (approved):
Thanks so much! 🙌

PR #230, opened 2023-08-09
Mock `rsconnect::writeManifest()` in test

juliasilge:
Recent releases of renv + rsconnect have caused some problems for this Connect test, since we can't make a manifest that is actually deployable with a source installation of vetiver. Let's see if this will fix it. 🤞


aronatkins (approved):
🌈 🥇

Issue #256, opened 2023-11-02
vetiver_sm_build . Error in enc2utf8(data) : argument is not a character vector   


joscani:
Hi. I've created a role with all sagemaker permissions . And I try to write docker using vetiver_sm_build. I've created an Ecr registry and in fact I've update a docker (not related with vetiver model) and use this docker in app runner. 
I've read this issue https://github.com/rstudio/vetiver-r/issues/215 but I have problems. 
My problem is with this when I try to use this from my local computer.
My ecr is joscani-irlanda
```r

execution_role <- "arn:aws:iam::969140037785:role/sagemaker-vetiver"

name = "bt_xgboost"


# Sends dockerfile to AWS Codebuild for docker build
image_uri <- vetiver_sm_build(
  board =  board,
  name = "bt_xgboost",
  # predict_args = predict_args,
  # docker_args = docker_args,
  repository = glue::glue("joscani_irlanda-{name}:{strftime(Sys.time(), '%Y-%m-%d')}") ,
  bucket = board$bucket,
  role = execution_role
)
- The lockfile is already up to date.  
Error in enc2utf8(data) : argument is not a character vector     

```

The role has been created using sagemakerfullaccess policy. Any idea. Thanks

juliasilge:
I am predicting that you are seeing this error from [here in the paws package](https://github.com/paws-r/paws/blob/035cbc64d619a61cb6528e9abc3dccad5c44bf5f/paws.common/R/signer_v4.R#L446), which is building up a signature for an AWS service when it is called. Both the S3 backend for the pins board and the smdocker R package use paws, so that doesn't help narrow it down a whole ton.

Can you walk through [what `vetiver_sm_build()` does internally, step by step,](https://github.com/rstudio/vetiver-r/blob/82f08aac9bc76c4152b8646c4eb4824cbdb6007d/R/sagemaker.R#L188) and find out in more detail where the error is happening? If it is during `vetiver_prepare_docker()` I can definitely help debug but if it is from `smdocker::sm_build()` I may need to point you to that repo instead.

stefanlinner:
Hello, 

my guess would be that the repository name is not valid. In AWS I experienced that I'm not allowed to use "_" (underscores) in names. So maybe try `repository = glue::glue("joscani-irlanda-{name}:{strftime(Sys.time(), '%Y-%m-%d')}")` instead. Maybe that already fixes your issue.

[...]

Issue #253, opened 2023-10-30
`vetiver_sm_endpoint()` error "object b not found"


stefanlinner:
Hi everyone, 

Thanks a lot for your work on this package!

I try to deploy a vetiver API in SageMaker from my local RStudio (not within SageMaker). Following the procedure discussed in this issue #215, I managed to set up my ARN IAM roles accordingly so that `vetiver_sm_build()` and `vetiver_sm_model()` run without any further problems. However, I am somehow failing to set up an API endpoint with `vetiver_sm_endpoint()`.

I am running the following code where I am using an execution role setup according to your explanations above and also my aws_acces_key_id and aws_secret_access_key are provided in .Renviron.

```r
library(vetiver)
library(pins)

name <- "cars-lm"

lm_model <- lm(mpg ~ ., data = mtcars)

v_model <-
  vetiver_model(
    lm_model,
    name
  )

model_board <-
  board_s3(
    bucket = "mlr4all-test-bucket",
    region = "us-east-1"
  )

model_board %>%
  vetiver_pin_write(v_model)

execution_role <- "arn:aws:iam::069650456288:role/vetivertest"

new_image_uri <-
  vetiver_sm_build(
    board = model_board,
    name = name,
    repository = glue::glue("vetiver-sagemaker-{name}:{strftime(Sys.time(), '%Y-%m-%d')}"),
    role = execution_role,
    bucket = model_board$bucket
  )

model_name <-
  vetiver_sm_model(
    image_uri = new_image_uri,
    role = execution_role
  )

new_endpoint <-
  vetiver_sm_endpoint(
    model_name = model_name,
    instance_type = "ml.t2.medium"
  )
```

Everything works smoothly until `vetiver_sm_endpoint()`. There I get the following error message:

<img width="613" alt="error_message" src="https://github.com/rstudio/vetiver-r/assets/83122367/f3a73162-2e5f-4be1-940c-44c97680386f">

When I check CloudWatch for the respective endpoint I find the following: 

<img width="693" alt="error_cloud_watch" src="https://github.com/rstudio/vetiver-r/assets/83122367/25e78d1c-1e24-4fe5-8581-dc0695952cd3">

I don't really understand that error message and I also don't know where this error could come from. 

Any help would be really nice!
Thank you guys a lot for your work!

<details style="margin-bottom:10px;">
<summary>
Session info
</summary>

``` r
sessioninfo::session_info()
#> ─ Session info ───────────────────────────────────────────────────────────────
[...]

PR #241, opened 2023-09-07
Use base R `deparse()` when writing `plumber.R`

juliasilge:
Closes #239 

With this change, I now can correctly generate code for a `board_url()`.

``` r
library(pins)
b <- board_url(c(
  chicago_model = 'https://colorado.posit.co/rsc/chicago-rstats-model-pin/'
))

deparse(board_deparse(b))
#> [1] "board_url(c(chicago_model = \"https://colorado.posit.co/rsc/chicago-rstats-model-pin/\"))"
```

<sup>Created on 2023-09-07 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>

PR #257, opened 2023-11-07
Go back to `rlang::expr_deparse()` except for `board_url()`

juliasilge:
Addresses these user-reported issues:

- #253 
- #256 

The change in #241 turned out to actually break generating the plumber files for more complex boards like S3 and similar. 😩 

I need to think about testing across different board types in vetiver. The pins package is tested very extensively across all the backends but vetiver does not, and here it is the code that _vetiver_ ran that was the problem. I could probably get away with mocking `vetiver_pin_read()` in such tests to avoid needing to call S3 like pins CI does.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,7 @@
 # vetiver (development version)
 
+* Fixed bug in generating plumber files (#257).
+
 # vetiver 0.2.4
 
 * Fixed how plumber files are generated for `board_url()` (#241).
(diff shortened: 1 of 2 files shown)


juliasilge:
If I'm going to write some tests with mocking, I need to export `new_board()` from pins.

juliasilge:
I'm going to fix this bug now, and wait on the tests since I will need to do a pins release to write the tests.

PR #266, opened 2023-12-07
refactor: change task_prototype to data_prototype

be-marc:
We want to remove`$task_prototype` in mlr3 and use  `$data_prototype` instead. Both will be available for some time.

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -55,7 +55,7 @@ Suggests:
     LiblineaR,
     luz,
     mgcv,
-    mlr3,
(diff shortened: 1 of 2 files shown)


juliasilge:
Thanks so much for this PR! 🙌 

Can we set a version requirement for mlr3 in `DESCRIPTION` for when this new `data_prototype` is available?

be-marc:
I set to our latest version. Thanks!

juliasilge (approved):
Thank you @be-marc! 🙏

Issue #219, opened 2023-06-09
Consider tools for extending vetiver to custom models


juliasilge:
I think you're right that we should consider making it easier for folks to extend vetiver to custom models. In Python, this [already works](https://rstudio.github.io/vetiver-python/stable/custom_handler.html) because of how the handlers are set up. In R, we are using S3 methods so it's not as straightforward right now. In the longer term, we could think about how to expose an easier way to support custom models. In the shorter term, it's worth documenting what you need to do to extend vetiver for now. Some similar documentation exists for [butcher](https://butcher.tidymodels.org/articles/adding-models-to-butcher.html) and [broom](https://www.tidymodels.org/learn/develop/broom/). Maybe we could make something like butcher's `new_model_butcher()` that generates a file with example code to start from.

_Originally posted by @juliasilge in https://github.com/rstudio/vetiver-r/issues/214#issuecomment-1584774130_

PR #271, opened 2024-02-28
New `additional_pkgs` argument for writing Plumber file

### LINKED ISSUE #270: Allow Custom Models to Assign Packages
## Background

This issue is in reference to the discussion had on #214.

Our team has a custom package responsible for making a model, and we want to set up and deploy a vetiver API endpoint from that same package. When working with the custom model wrapped by `vetiver_model()`, we had difficulty getting our deployed API on Posit Connect to correctly find the new methods we had made (i.e. `handler_predict.my_model`, etc.) and would get the error:

```
There is no method available to build a prediction handler for `x`.
```

## Current Output

Indicating those methods weren't being carried over. The default output file for `vetiver::vetiver_write_plumber()` looks something like:

```r
# Generated by the vetiver package; edit with care

library(pins)
library(plumber)
library(rapidoc)
library(vetiver)

# Packages needed to generate model predictions
if (FALSE) {
    library(myCustomPackage)
    library(otherusefulpackages)
    ...
}
b <- board_connect(auth = "envvar")
v <- vetiver_pin_read(b, "my_model")

#* @plumber
function(pr) {
    pr %>% vetiver_api(v)
}

```

What we found worked was moving the `library()` call for the custom package out of `if (FALSE) { }`. Then the deployed API correctly picked up the custom methods for our model.

## Proposed Solution

It might be helpful to allow users to edit what packages are attached to the default plumber output, currently they are rigidly set in `infra_pkgs`:

https://github.com/rstudio/vetiver-r/blob/581a4e98d9673013a386a9715f180f025fc3f03f/R/write-plumber.R#L96

juliasilge:
Closes #270 

This PR exposes a new optional argument for `vetiver_write_plumber()` that lets folks provide packages that need to be attached (like with `library()`) to run the API. I don't know why I was confused about this before, because OF COURSE folks will need to be able to attach some custom package here if they have used it to extend vetiver. This work is necessary to fully support #219 in the future.


juliasilge:
@rsh512 would you be up for installing this via `devtools::install_github("rstudio/vetiver-r@extra-infra-pkgs")` to see if this solves your problem? If you would normally use `vetiver_deploy_rsconnect()`, you can do it like:

```r
vetiver_deploy_rsconnect(
    board,
    "user.name/model-name",
    additional_pkgs = "yourcustompkg"
)
```

R/write-plumber.R
```diff
@@ -12,6 +12,8 @@
 #' working directory. See [plumber::plumb()] for naming precedence rules.
 #' @param rsconnect Create a Plumber file with features needed for [Posit
 #' Connect](https://posit.co/products/enterprise/connect/)? Defaults to `TRUE`.
+#' @param additional_pkgs Any additional R packages that need to be **attached**
```
juliasilge, line 15: I called this `additional_pkgs` just like the argument in `vetiver_write_docker()` 🤪 because I thought it would be worse to give it a different name, even though they are different kinds of extra packages needed (one set needs to be just installed and one set needs to be fully attached). @isabelizimm do you have a different opinion here?

R/rsconnect.R
```diff
@@ -55,15 +55,17 @@
 vetiver_deploy_rsconnect <- function(board, name, version = NULL,
                                      predict_args = list(),
                                      appTitle = glue::glue("{name} model API"),
-                                     ...) {
+                                     ...,
+                                     additional_pkgs = character(0)) {
```
juliasilge, line 59: I piped this new arg through to the Connect functions, but I did _not_ pipe it through for `vetiver_prepare_docker()`, since this is more like needing to customize the Docker `additional_pkgs` which we don't let you do there. In that situation, we say "use the more modular functions" instead of the higher level one.

R/write-plumber.R
juliasilge: There is no way to accidentally send what you mean to use for one arg to the other one, but maybe this is too confusing for people to understand easily. I could make them `additional_docker_pkgs` and `additional_plumber_pkgs`, for more clarity. Or `additional_installed_pkgs` and `additional_attached_pkgs`. This would involve deprecating the existing `additional_pkgs` argument.

rsh52:
> @rsh52 would you be up for installing this via `devtools::install_github("rstudio/vetiver-r@extra-infra-pkgs")` to see if this solves your problem? If you would normally use `vetiver_deploy_rsconnect()`, you can do it like:
> 
> ```r
> vetiver_deploy_rsconnect(
>     board,
>     "user.name/model-name",
>     additional_pkgs = "yourcustompkg"
> )
> ```

Just tested `vetiver::vetiver_deploy_rsconnect()` / `vetiver::vetiver_write_plumber()` and successfully deployed without a hitch! `additional_pkgs` is exactly what we were looking for, thank you!

R/write-plumber.R
isabelizimm: I think `additional_pkgs` seems fine-- I don't think we want to try to differentiate between docker/plumber/etc packages, since you will be installing all of them at the endpoint regardless, right?

juliasilge:
So glad to hear this @rsh52! 🙌 

I'm trying to handle the CI failures in #272 and then will merge that into this PR.

PR #272, opened 2024-02-28
Some boring maintenance, to get CI passing

juliasilge:
We've got some new failures in CI to address.


juliasilge:
Closes #273 

Thank you SO VERY MUCH @dfalbel! 🙏

PR #287, opened 2024-07-19
When a recipe has a `ptype`, use it to set up the vetiver model

### LINKED ISSUE #286: recipes now support ptype extraction
This will help with validations when using a recipe by itself in vetiver

1.1.0 included `recipes_ptype()` and `recipes_ptype_validate()`

older recipes that doesn't have information will return `NULL`

juliasilge:
Closes #286 

~~The recipes tests now skip on older versions but I didn't bump in `DESCRIPTION`. Maybe I should just bump though, actually.~~

I did decide to bump the version for recipes in `DESCRIPTION`; I think that's better than trying to support both moving forward.

### DIFF
--- a/R/recipe.R
+++ b/R/recipe.R
@@ -16,11 +16,17 @@ vetiver_create_meta.recipe <- function(model, metadata) {
 #' @rdname vetiver_create_ptype
 #' @export
 vetiver_ptype.recipe <- function(model, ...) {
-    rlang::check_dots_used()
-    dots <- list(...)
-    check_ptype_data(dots)
-    ptype <- vctrs::vec_ptype(dots$prototype_data)
-    tibble::as_tibble(ptype)
+    ptype <- model$ptype
+    if (is_null(ptype)) {
+        rlang::check_dots_used()
+        dots <- list(...)
+        check_ptype_data(dots)
+        ptype <- vctrs::vec_ptype(dots$prototype_data)
+        return(tibble::as_tibble(ptype))
+    }
+    preds <- vec_slice(model$var_info, model$var_info$role == "predictor")
+    preds <- preds$variable
+    ptype[preds]
 }
 
 #' @rdname vetiver_create_description
--- a/R/recipe.R
+++ b/R/recipe.R
@@ -16,8 +16,7 @@ vetiver_create_meta.recipe <- function(model, metadata) {
 #' @rdname vetiver_create_ptype
 #' @export
 vetiver_ptype.recipe <- function(model, ...) {
-    preds <- vec_slice(model$var_info, model$var_info$role == "predictor")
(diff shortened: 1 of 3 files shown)


juliasilge:
@EmilHvitfeldt Let me know if you have any opinions on this!

R/recipe.R
```diff
@@ -16,11 +16,8 @@ vetiver_create_meta.recipe <- function(model, metadata) {
 #' @rdname vetiver_create_ptype
 #' @export
 vetiver_ptype.recipe <- function(model, ...) {
-    rlang::check_dots_used()
-    dots <- list(...)
-    check_ptype_data(dots)
-    ptype <- vctrs::vec_ptype(dots$prototype_data)
-    tibble::as_tibble(ptype)
+    preds <- vec_slice(model$var_info, model$var_info$role == "predictor")
+    model$ptype[preds$variable]
```
EmilHvitfeldt, line 20: This is not in line with what recipes expects at bake time. As per [?update_role_requirements()](https://recipes.tidymodels.org/reference/update_role_requirements.html)

The following represents the default bake time requirements of specific types of roles:

- `"outcome"`: Not required at bake time. Can't be changed.

- `"predictor"`: Required at bake time. Can't be changed.

- `"case_weights"`: Not required at bake time by default.

- `NA`: Required at bake time by default.

- Custom roles: Required at bake time by default.

The above code fails for `NA` roles and custom roles.

``` r
data_orig <- tibble::tibble(
  y = 1:10,
  id = 1:10,
  x1 = letters[1:10],
  x2 = factor(letters[1:10]),
  cw = hardhat::importance_weights(1:10)
)

library(recipes)

model <- recipe(data_orig) |>
  update_role(y, new_role = "outcome") |>
  update_role(x1, x2, new_role = "predictor") |>
  update_role(id, new_role = "keep_at_bake") |>
  prep()

library(vctrs)

preds <- vec_slice(model$var_info, model$var_info$role == "predictor")

model$ptype[preds$variable]
#> # A tibble: 0 × 2
#> # ℹ 2 variables: x1 <chr>, x2 <fct>

recipes::recipes_ptype(model, stage = "bake")
#> # A tibble: 0 × 3
#> # ℹ 3 variables: id <int>, x1 <chr>, x2 <fct>
```

I would recommend that you use `recipes::recipes_ptype()` instead
```suggestion
    recipes::recipes_ptype(model, stage = "bake")
```

EmilHvitfeldt (approved):
so happy to see recipes ptype in use!

R/recipe.R
juliasilge: Right right right, I am remembering now. Thanks!

EmilHvitfeldt: i only remember because we have it written down 😅

PR #290, opened 2024-10-09
Add xgboost check

mfansler:
Adds a skip when `xgboost` is not installed. Otherwise, test will fail with:

```r
── Error ('test-mlr3.R:34:5'): learners from mlr3learners can be pinned ────────
<packageNotFoundError/error/condition>
Error: The following packages could not be loaded: xgboost
Backtrace:
    ▆
 1. └─learner$train(task) at test-mlr3.R:34:5
 2.   └─mlr3:::.__Learner__train(...)
 3.     └─mlr3:::learner_train(...)
 4.       └─mlr3misc::require_namespaces(learner$packages)
```


juliasilge (approved):
Thank you so much!

PR #293, opened 2025-09-25
plumber2 support

thomasp85:
This is my first attempt at providing plumber2 support in a way that feels native

The main functionality is captured by the `api_vetiver()` and the accompanying `@vetiver` annotation tag.

The `api_vetiver()` function is basically similar to `vetiver_api()` but follows the naming convention of plumber2 (`api_` prefixed), and of course works with plumber2 input instead plumber input.

The novelty is the `@vetiver` tag that allows you to mix in vetiver model serving within a plumber file. The syntax is currently:

```r
#* @vetiver /model/predict
vetiver_model(<some_model>, ...)
```

So you provide the path after the tag, and then below where you'd usually have a function you'd provide the vetiver model object however you like

This PR requires the dev version of plumber2 due to some bugs I uncovered

### DIFF
--- a/.github/workflows/R-CMD-check.yaml
+++ b/.github/workflows/R-CMD-check.yaml
@@ -35,6 +35,7 @@ jobs:
       GITHUB_PAT: ${{ secrets.GITHUB_TOKEN }}
       R_KEEP_PKG_SOURCE: yes
       TORCH_INSTALL: 1
+      PY_REQUIRE_LEGACY_KERAS: 1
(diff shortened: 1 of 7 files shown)


thomasp85:
One thing that could be nice is if it was possible to statically determine the output format of the prediction so that it could be added to the OpenAPI documentation, but I don't assume that is possible? Is dummy data included in the object so it can be predicted upon and the output inspected?

PR #304, opened 2025-12-08
Xgboost fixes

EmilHvitfeldt:


EmilHvitfeldt:
please verify. but I think the GHA errors are only coming from mlr3, which are also trying to deal with the xgboost changes https://github.com/mlr-org/mlr3learners/pull/366

